PR #7, opened 2017-08-08
Compcert submodule

vzaliva:
Instead of asking the user to download CompCert manually it was added as sub-module into the project. Now if the project is cloned with --recursive flag, CompCert will be automatically extracted from it's official Git repository into lib/CompCert.


vzaliva:
Submodules tutotrial: https://git-scm.com/book/en/v2/Git-Tools-Submodules

PR #44, opened 2018-06-08
Extend the Exception Monad with a catch operation. Axiomatize its laws

### LINKED ISSUE #42: ExceptionMonad is lawless, add laws?
The current definition of [`ExceptionMonad`](https://github.com/vellvm/vellvm/blob/8ba75559303f1cf28b5466c17640c2e4574a0aa5/src/coq/Classes.v#L422) has no laws associated to this theory. I would have excepted to find a law:

```coq
raise e >>= f ≡ raise e.
```

to reason about error propagation.

Will there be the functionality of `catchError` as provided by [`MonadError`](http://hackage.haskell.org/package/mtl-2.2.2/docs/Control-Monad-Except.html)?

In that case, we would need another law:

```hs
catchError (raise e) f ≡ f  e
```
This blog post mentions one more rule about the interaction between [`ret` and catching](https://lukajcb.github.io/blog/functional/2018/04/15/rethinking-monaderror.html).

Unfortunately, I can't find an authoritative source on laws off-hand.

YaZko:
The Exception Monad is currently only used as an Error one, and hence was minimalist. This PR adds a catch operator and axiomatizes a few of its laws.

Closes #42

### DIFF
--- a/src/coq/Classes.v
+++ b/src/coq/Classes.v
@@ -419,10 +419,33 @@ Defined.
 
 (* Monads with error -------------------------------------------------------- *)
 
-Class ExceptionMonad E F `{Monad F} := raise : forall {A}, E -> F A.
+Class ExceptionMonad E F `{Monad F} :=
+  { raise : forall {A}, E -> F A;
+    catch : forall {A}, F A -> (E -> F A) -> F A }
+.
+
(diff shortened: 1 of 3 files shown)


bollu:
Thank you :) This works perfectly.

PR #46, opened 2018-06-12
Add theory and induction schemes to make equality decidable on `typ`.

bollu:
To show that equality is decidable on `typ`, we need a stronger
induction principle than the one generated by default from Coq.

Use the trick/folklore in CPDT from the "Nested Inductive type" section to
generate a stronger induction principle.

TODO: Extend this to show decidable equality of `exp`, and in general
all of `LLVMAst`.

This is useful to perform pattern matching on the IR.


Zdancewic:
This old pull request (thanks!) was subsumed by the DecidableEquality.v file.

PR #48, opened 2018-06-13
Generate documentation using coq2html?

bollu:
Here's a PR in case `vellvm` wishes to use  `coq2html` to generate documentation. I personally like the output better than `coqdoc`.

EDIT: [Link to sample `Classes.v` documentation](https://bollu.github.io/vellvm/src/doc/html/Classes.html)

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -81,4 +81,18 @@ clean:
 	rm -rf output
 	rm -f vellvm
 
+doc/coq2html: doc/coq2html.ml
+	ocamlopt -w +a-29 -o doc/coq2html str.cmxa doc/coq2html.ml
+
+doc/coq2html.ml: doc/coq2html.mll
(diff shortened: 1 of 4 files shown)


Zdancewic:
This is a good idea.  Do you know if there's a way to generate the new Sphinx-ified documentation in the style of the reference manual?

Zdancewic:
It looks like we need to add coq2html.css and coq2html.js to the doc directory.  Perhaps we should add coq2html as a git submodule in /lib (just like Paco) and fix the makefiles to assume that?

bollu:
Yes, that makes sense. Sorry for not testing this properly, I must had this locally, but I didn't commit the right files...

bollu:
[Here is the docs](https://github.com/coq/coq/tree/master/doc/sphinx#documenting-coq-with-sphinx) on how Coq generates sphinx docs.

PR #56, opened 2018-06-27
add coq2html as a submodule

bollu:
Trying to add `coq2html` as a submodule so we can build docs correctly. Please do verify before pulling this in :)


Zdancewic:
This works great.  Thanks!  (Note to self: need to do `git submodule update --recursive` after adding a new submodule.)

Issue #57, opened 2018-06-27
Return non-zero exit code when vellvm --test fails?


bollu:
I'm trying to setup travis, and this incorrectly succeeds, even though there is a failing vellvm test:
```
╭─bollu@cantordust ~/work/vellvm/origin/src  ‹4e16716*› 
╰─$ ./vellvm --test; echo $?
(* -------- Vellvm Test Harness -------- *)
Running llc on: ../tests/ll/sum_tree.ll
Running llc on: ../tests/ll/global2.ll
Running llc on: ../tests/ll/bitcast1.ll
Running llc on: ../tests/ll/puts.ll
Running llc on: ../tests/ll/br2.ll
Running llc on: ../tests/ll/global1.ll
Running llc on: ../tests/ll/args1.ll
Running llc on: ../tests/ll/binary_gcd.ll
Running llc on: ../tests/ll/call5.ll
Running llc on: ../tests/ll/cbr1.ll
Running llc on: ../tests/ll/alloca2.ll
Running llc on: ../tests/ll/alloca1.ll
Running llc on: ../tests/ll/list.ll
Running llc on: ../tests/ll/return_intermediate.ll
Running llc on: ../tests/ll/ashr.ll
Running llc on: ../tests/ll/arith_combo.ll
Running llc on: ../tests/ll/matmul.ll
Running llc on: ../tests/ll/callback1.ll
Running llc on: ../tests/ll/gep8.ll
Running llc on: ../tests/ll/global5.ll
Running llc on: ../tests/ll/qtree.ll
Running llc on: ../tests/ll/return.ll
Running llc on: ../tests/ll/call4.ll
Running llc on: ../tests/ll/gep5.ll
Running llc on: ../tests/ll/helloworld.ll
Running llc on: ../tests/ll/cbr2.ll
Running llc on: ../tests/ll/gep1.ll
Running llc on: ../tests/ll/global4.ll
Running llc on: ../tests/ll/gep_simple.ll
Running llc on: ../tests/ll/and.ll
Running llc on: ../tests/ll/sub_neg.ll
Running llc on: ../tests/ll/naive_factor_nonprime.ll
Running llc on: ../tests/ll/aggregate-mem.ll
Running llc on: ../tests/ll/scope.ll
Running llc on: ../tests/ll/gep7.ll
Running llc on: ../tests/ll/factrect.ll
Running llc on: ../tests/ll/or.ll
Running llc on: ../tests/ll/lfsr.ll
[...]

PR #58, opened 2018-06-27
[WIP] Setup Travis CI

### LINKED ISSUE #55: Setup CI?
I see there is already a [travis branch](https://github.com/vellvm/vellvm/tree/travis). Could this be merged? That would help validate all pulls.

bollu:
The CI infrastructure I setup is very dumb right now, we can do much better at multiple places:

1. don't build coq each time. Rather, pull from coq packages (which contains nightlies as well), which are packaged by @jasongross. [`fiat-crypto`](https://github.com/mit-plv/fiat-crypto/blob/master/.travis.yml#L20) does this.

2. Find some way to not have to recompile `CompCert` each time. Is there some way to cache submodule builds? I'm not sure. A quick google search does not show anything related (there are threads on caching submodule clones, like [this Rust issue](https://github.com/rust-lang/rust/issues/40772) and [this commit on `qemu`](https://patchwork.kernel.org/patch/9803371/), but none on caching submodule builds).

2.b Possibly, we can extract out the `CompCert` files `vellvm` needs? I think `vellvm` only depends on `Integers`, `Floats`, and maybe some other odds-and-ends, correct? So there is no need to build _all_ of `CompCert`.

3. Similarly, find some way to not have to recompile `paco` every time.

There is also the (blocking) issue that the test harness failure does not return a non-zero exit code. [I filed an issue](https://github.com/vellvm/vellvm/issues/57). Once this is fixed, a test harness failure will cause Travis to log the build as `failed`.

#### Example run:
[here's a link to an example run](https://travis-ci.org/bollu/vellvm/builds/397379318)
#### Blocking issues
- [x] teach test harness to return non-zero exit code (Fixed by [`0fbf3e2`](https://github.com/vellvm/vellvm/commit/0fbf3e2c177fdb1300bb4107f4e194a3f46f4895))

Closes #55

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,53 @@
+dist: trusty
+sudo: required
+language: c
+cache:
+  apt: true
+  directories:
+  - $HOME/.opam
+addons:
+  apt:
+    sources:
+    - avsm
+    packages:
+    - opam
+    - aspcud
+env:
+  global:
+  - NJOBS=1
+  # https://github.com/AbsInt/CompCert/issues/234
+  - EXTRA_OPAM=menhir.20180528
+
+  # for CompCert: http://compcert.inria.fr/man/manual002.html
+  - COMPILER="system"
+  # Main test targets
+  # get versions from: https://opam.ocaml.org/packages/coq/
+  # Only use versions that CompCert can build with...
+  matrix:
+  - TEST_TARGET="8.6.1"
+  - TEST_TARGET="8.7.2"
+  - TEST_TARGET="8.8.0"
+
+before_script:
+  -  ./ci/keep_alive.sh &
+
+install:
+  - opam init -j ${NJOBS} --compiler=${COMPILER} -n -y
+  - eval $(opam config env)
+  - opam config var root
+  # this somehow fails the second time...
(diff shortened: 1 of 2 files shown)


bollu:
Rebased over (0fbf3e2)[https://github.com/vellvm/vellvm/commit/0fbf3e2c177fdb1300bb4107f4e194a3f46f4895], which fixed exit-code values. 

Check that the [current travis build](https://travis-ci.org/bollu/vellvm/builds/397403638) fails _at the correct point_.

Then, cleanup branch!

bollu:
Great, it appears that all the tests in the build [fail properly](https://travis-ci.org/bollu/vellvm/builds/397403638), i.e, they build `vellvm`, run the tests, and then fail.

bollu:
[Travis build](https://travis-ci.org/bollu/vellvm/builds/397417724) of rebased version link.

bollu:
So, the tests fail as expected. I think this can be merged.

Are there more things I need to fix?

I'd like this to be merged if it is OK, so we can iterate on this.

JasonGross:
If you want to cache submodule builds, you might be able to copy what I do in `fiat-crypto` to cache build-products so that I can build in multiple stages (to get more time), and have Coq pick up where I left off.

PR #62, opened 2018-09-27
Added a typeclass for VELLVM integers.

Chobbes:
Currently this just wraps the CompCert types so we don't need to duplicate as much complex code.

Can probably clean this up still. Had to update some of the Lemmas in Renaming.v... Can also get rid of integer_op if we can get rid of it in Renaming.v. Not sure how much this affects the renaming proof.

### DIFF
--- a/src/coq/DynamicValues.v
+++ b/src/coq/DynamicValues.v
@@ -379,11 +379,15 @@ Class VInt I : Type :=
   (* Arithmetic Operations ---------------------------------------------------- *)
   Section ARITHMETIC.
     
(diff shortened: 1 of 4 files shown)


Chobbes:
Oh good point! I'll go through and fix that when I have the chance!

I want to look into cleaning up the conversions in StepSemantics.v more as well. I need to think about it more, though.

Chobbes:
@Grain this is what I was mentioning that touched renaming slightly. I'm not sure that it will matter for you, but there were a lot of eval_intXX_op cases in there that kind of just go away.

Issue #64, opened 2018-10-24
use coq-paco


vzaliva:
lib/paco is no longer needed as `coq-paco` opam package was updated to recent version and should be used instead. See also https://github.com/coq/opam-coq-archive/pull/260

Zdancewic:
done!

PR #65, opened 2018-10-24
removed paco submodile, using coq-paco OPAM package instead 

vzaliva:
https://github.com/vellvm/vellvm/issues/64

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,6 @@
 [submodule "lib/CompCert"]
 	path = lib/CompCert
 	url = https://github.com/AbsInt/CompCert.git
-[submodule "lib/paco"]
-	path = lib/paco
-	url = https://github.com/vzaliva/paco
 [submodule "lib/coq2html"]
 	path = lib/coq2html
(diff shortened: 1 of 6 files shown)


vzaliva:
I've tested with coq-8.8.1 and 8.8.2 using opam1 and opam2 respectively. `coq-paco` package seems to contain the most recent version and compatible with both.

As to opam2, in general, I am also struggling with the decision. I feel that moving to opam2 is the right step, but the main thing stopping me is the fact that there is no official package for Ubuntu. I already switched to opam2 on my mac machine which is using homebrew and it works without the problem so far and installs newer versions of several packages like dune, coq. On my main development machine, which is Ubuntu I am still on opam1 but planning to switch soon, using this https://discuss.ocaml.org/t/opam-2-0-experimental-ppas/2446

vzaliva:
the reason Travis test fails on this pull request is because it is trying to build `lib/paco`. This step should be removed from the build process.

vzaliva:
OK, I found and fixed travis config. CI tests pass now

Zdancewic:
Thanks!  This looks great.

PR #66, opened 2018-10-25
switch to `dune` as build system instead of `ocamlbuild`

vzaliva:
Dune is arguably a more capable and modern build system for OCaml projects and I suggest to switch to it for OCaml builds.

An additional motivation for this changes is to be able to include VELLVM in other projects as a submodule. Since dune builds are composable, it will be easier to build it as a part other project builds.

Tested with the following configurations:
  1. dune-1.2.1, coq-8.8.1, and opam-1.2
  1. dune-1.4.0, coq-8.8.2, and opam-2.0.1

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ src/ml/extracted/*.ml
 src/ml/extracted/STAMP
 src/.depend
 output
+/src/ml/extracted/.merlin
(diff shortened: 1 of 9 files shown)


vzaliva:
another motivation for this is to be able to include vellvm in other project as submodule. Since dune builds are composable, it will be easier to build it as a part other project builds.

Zdancewic:
This works great, thanks!

Issue #71, opened 2018-11-21
Proposal: comments in AST


vzaliva:
When generating IR code via Vellvm AST it could be very useful to be able to include comments. Such comments will not affect semantics but will be passed through to pretty-printed IR code. It will make generated code much more readable and easy to debug.

I can implement such an extension, but before doing so would like to check if there are any objections or thoughts on this.

Zdancewic:
This is a good idea.  I don't think that it will be too hard to add this feature.  We will want to make sure that doing so doesn't add too much overhead to proofs, but that should be pretty easy by ignoring the comments for the purposes of the semantics.

vzaliva:
merged

PR #72, opened 2018-11-23
Comments

vzaliva:
Per https://github.com/vellvm/vellvm/issues/71 this PR adds some places where comments could be inserted. In particular:

1. In top-level constructs (`TLE_Comment`)
2. In `code` section (`INSTR_Comment`)
3. In blocks, before the label (`blck_comments` record field)

I've tested it in my development which uses Vellvm and this works as expected and makes generated code easier to read and debug.

Comments added to AST and pretty-printed to IR. However, IR parsing does not add comments entries to AST.


Zdancewic:
Thanks!  This looks great.

PR #88, opened 2019-02-07
compatibility with coq-flocq 3.1.0

vzaliva:


Zdancewic:
I merged this update -- the only slight problem is the persistent issue with coq-floqc opam installation on OSX, but the solution of commenting out the sandboxing in ~/.opam/config worked.

PR #90, opened 2019-02-13
Refactor debug to be a separate effect from IO.

Chobbes:
Is this the right grouping of effects? The left associativity of `+'` seems a bit awkward, because each added effect buries the first (probably most important effect, like IO) deeper in sums, which means more changes? Hence the parenthesis... I don't know what's worse.


Lysxia:
I forgot to change the associativity! (Effect rows used to be snoc lists for no reason.)

Chobbes:
> I forgot to change the associativity! (Effect rows used to be snoc lists for no reason.)

Oh, awesome, so this will change in Itrees? :)

Lysxia:
Yeah it just changed!

Chobbes:
New change renames the Trace monad to the LLVM monad.

Looking into the predicate stuff now.

Zdancewic:
I ran into an issue where the ITree library defines monad notations that are also defined in ExtLib.  We want the more general ExtLib ones, so I import it second.  (This should probably be fixed in the ITrees library.)

PR #92, opened 2019-03-25
port to new itrees library

Zdancewic:


vzaliva:
I've tested this.

1. Default check out brings out an older version of itrees library (which still has Std.v). We probably need to update module reference to point to the newer version.
2. after doing:
```
cd lib/InteractionTrees/
git pull origin master
git checkout master
```

I was able to compile this branch and it passed all unit tests.

vzaliva:
@Zdancewic I too a liberty of amending this pull request with patch to point to newer version of iTress. Hopefully this will give a user the right version on checkout.

Zdancewic:
Thanks!  I've merged this to master.

Issue #95, opened 2019-04-25
float literals


vzaliva:
The problem is best illustrated by the following example:

```
llc -relocation-model=pic -asm-verbose -O0 test_binop_plus.ll
llc: test_dynwin32.ll:2:43: error: floating point constant invalid for type
@D = internal constant [3 x float] [float 76756783.5125, float -243791519.156, float -197089434.476], align 16
                                          ^
Makefile:22: recipe for target 'test_dynwin32.s' failed
```

This code is pretty-printed from LLVM AST. The problem is that `76756783.5125` could not be represented as an exact floating point value. (You can check here: https://www.h-schmidt.net/FloatConverter/IEEE754.html). The IR reference states:

> "The assembler requires the exact decimal value of a floating-point constant. For example, the assembler accepts 1.25 but rejects 1.3 because 1.3 is a repeating decimal in binary."

The question is how such constant appeared in the first place. The value in question was generated from `EXP_Float` with the following code:

`EXP_Float (Float.of_single b32)` where `b32` has type `binary32`

It was done as a workaround for https://github.com/vellvm/vellvm/issues/73

To sum it up, we started with a valid 32-bit floating point value, which we converted to 64-bit floating point constant. This constant was pretty-printed as an ASCII string which does not represent exact 32-bit floating point value.

One possible explanation (source: https://cs.stackexchange.com/questions/56976/whether-assigning-of-single-precision-ieee754-float-to-double-is-reversible ) is while the 64-bit value is the same as 32-bit, but when printed it is rendered to a different decimal string which could not be parsed back without precision loss. In other words:

- `binary32 -> binary64 -> binary32`  - **OK**
- `binary32 -> binary64 -> string -> binary32` -  **Precision loss!**

If my hypothesis is correct it could be fixed by either making sure 32-bit floats are properly pretty printed, or by revisiting https://github.com/vellvm/vellvm/issues/73 and using proper binary32 to store 32-bit floating point constants internally.

Zdancewic:
Interesting!  It seems like patching the pretty-printer is the right way to go.  

Unfortunately, this isn't quite so straightforward to fix as you might want because OCaml doesn't have a 32bit floating point value.  This means that we probably need to call out to a small C function that does the right thing.  We'll also have to add a bit of case analysis to the pretty printer to determine whether to print an EXP_Float value as either a single or double precision string representation.

We'll look into this over the next few days!

Zdancewic:
[...]

PR #96, opened 2019-04-26
Float print in hex

Chobbes:
Addressing issue #95. This might need a little bit of extra work.

Also this fixes the old hex printing (apparently %Lx doesn't work, you have to use %LX).

I encountered some oddities here. Described below.

    LangRef describes this here:

    https://llvm.org/docs/LangRef.html#simple-constants

    Initially I thought I would just take the hexadecimal representation of
    the float. I.e., I would write out a good 32-bits worth of hexadecimal. So, in
    our example 125.32 would become:

    0x42faa3d7

    (https://www.exploringbinary.com/floating-point-converter/)

    However, LLVM says:

    > When using the hexadecimal form, constants of types half, float, and
    > double are represented using the 16-digit form shown above (which
    > matches the IEEE754 representation for double); half and float values
    > must, however, be exactly representable as IEEE 754 half and single
    > precision, respectively.

    So, I *have* to use 64-bits worth of hexadecimal. Which... Fine? I
    thought it might be:

    0x0000000042faa3d7

    It's not. And it's also not:

    0x42faa3d700000000

    It's actually, according to a C program I wrote...

    0x405f547ae0000000

    Which makes sense because the double is:

    0x405f547ae147ae14

    So, they're lopping off the bottom 7 hex digits, leaving us with
    36-bits remaining. Why 36 and not 32? Because doubles have 11
    exponent bits, and floats have 8. So we actually have the mantisa
    shifted over by 3 bits (and there's an extra 0 bit because 35 isn't a multiple of 4).

Ultimately this conversion seems to work (it all makes sense now, but it wasn't obvious to me at first glance), but I'm wondering if we should explicitly lop off the bits for. It sounds like @vzaliva might be altering the `EXP_Float` values, so this might still cause problems unless we do that? It's an easy change to make, but if we don't need it, maybe that's better? Thoughts?

### DIFF
--- a/src/coq/AstLib.v
+++ b/src/coq/AstLib.v
@@ -398,7 +398,8 @@ Inductive value : Set :=
   Variable P : exp -> Prop.
   Hypothesis IH_Ident   : forall (id:ident), P ((EXP_Ident id)).
   Hypothesis IH_Integer : forall (x:int), P ((EXP_Integer x)).
-  Hypothesis IH_Float   : forall (f:float), P ((EXP_Float f)).
+  Hypothesis IH_Float   : forall (f:float32), P ((EXP_Float f)).
+  Hypothesis IH_Double  : forall (f:float), P ((EXP_Double f)).  
   Hypothesis IH_Hex     : forall (h:float), P ((EXP_Hex h)).  
   Hypothesis IH_Bool    : forall (b:bool), P ((EXP_Bool b)).
   Hypothesis IH_Null    : P ((EXP_Null )).
@@ -431,6 +432,7 @@ Inductive value : Set :=
     - apply IH_Ident.
     - apply IH_Integer.
     - apply IH_Float.
+    - apply IH_Double.      
     - apply IH_Hex.      
     - apply IH_Bool.
     - apply IH_Null.
--- a/src/coq/CFGProp.v
+++ b/src/coq/CFGProp.v
@@ -155,7 +155,8 @@ Fixpoint exp_uses (v:exp) : list ident :=
   match v with
   | EXP_Ident id => [id]
   | EXP_Integer _
-  | EXP_Float _ 
+  | EXP_Float _
+  | EXP_Double _               
   | EXP_Hex _        
   | EXP_Bool _
   | EXP_Null
--- a/src/coq/LLVMAst.v
+++ b/src/coq/LLVMAst.v
@@ -31,7 +31,8 @@ Open Scope string_scope.
 Open Scope list_scope.
 
 Definition int := Z.
-Definition float := Floats.float.  (* -bit floating point value *)
+Definition float := Floats.float.  (* 64-bit floating point value *)
+Definition float32 := Floats.float32.
 
 Inductive linkage : Set :=
 | LINKAGE_Private
@@ -191,17 +192,22 @@ Definition tident : Set := (typ * ident)%type.
 (* NOTES: 
   This datatype is more permissive than legal in LLVM:
      - it allows identifiers to appear nested inside of "constant expressions"
+       that is OK as long as we validate the syntax as "well-formed" before 
+       trying to give it semantics
 
   NOTES:
    - Integer expressions: llc parses large integer exps and converts them to some 
      internal form (based on integer size?)
    
    - Float constants: these are always parsed as 64-bit representable floats 
-     using ocamls float_of_string function.  If they are used in LLVM as 32-bit 
-     rather than 64-bit floats, they are converted when evaluated.
+     using ocamls float_of_string function. The parser converts float literals 
+     to 32-bit values using the type information available in the syntax.
+
+     -- TODO: 128-bit, 16-bit, other float formats?
 
    - Hex constants: these are always parsed as 0x<16-digit> 64-bit exps and
-     bit-converted to ocaml's 64-bit float representation.
+     bit-converted to ocaml's 64-bit float representation.  If they are
+     evaluated at 32-bit float types, they are converted during evaluation.
 
    - EXP_ prefix denotes syntax that LLVM calls a "value"
    - OP_  prefix denotes syntax that requires further evaluation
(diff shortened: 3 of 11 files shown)


Chobbes:
Also this revealed an issue in the pretty print tests... If the format string is wrong, it will actually print out a blank file... Which is valid LLVM, so the tests don't actually catch any problems!

Zdancewic:
Does this actually address Vadim's original problem?  I would have expected that we need to do case analysis on the type of the `EXP_float` to determine whether to print it as a 64-bit literal or a 32-bit literal.  The information we need is available at the `texp` level of the syntax, but not at the `exp` level, right?

Chobbes:
> The code above will emit "0x0" for float value 0.0. Perhaps "0x%016LX" format should be used.

This is a good catch, I think you are right about this! Also unclear about the previous endianness comment. Did you look into this further? Is it a non-issue?

> Does this actually address Vadim's original problem? I would have expected that we need to do case analysis on the type of the EXP_float to determine whether to print it as a 64-bit literal or a 32-bit literal. The information we need is available at the texp level of the syntax, but not at the exp level, right?

This is what I'm a bit unclear about. Initially I went this route when I was trying to get this working, but it seemed like it was unnecessary to me --- it's easy enough to do this if we need it, but my understanding of the problem was that while doubles can store all single precision values accurately, when printing the decimal version of the number it prints something different than if you were to print the equivalent single precision value, and what it prints might not be representable as a single precision float.

So, for instance, this program uses the exact single precision floating point value:

> define float @main(i8 %argc, i8** %arcv) {
>   ret float 125.31999969482421875
> }

But when run through vellvm with test-pp-dir it spits out a file with a rounded string which leads to this problem:

> llc: error: llc: output/float_literal_1.v.ll:5:13: error: floating point constant invalid for type
>   ret float 125.319999695

This is what I thought the problem was, but maybe there's another problem as well. Anyway, with this change this problem is addressed and you get the expected hex value:

> define float @main(i8 %argc, i8** %arcv) {
> ; <label> 0
>   ret float 0x405F547AE0000000
> }

I might be misunderstanding something (so please correct me if I'm wrong), but I think @vzaliva is constructing the values with `EXP_Float (Float.of_single b32)`, so the value stored in the double should be precisely representable as a 32-bit float, right?

It's also reasonable that we might want to be more careful about this, because perhaps we want to make sure that we don't do a program transformation that will end up converting a 32-bit float literal into one that's only representable as a double. E.g., maybe we do constant propogation with floats, and say substitute a single literal for the result of the division between two float literals, but since the arithmetic is done with doubles perhaps the result is not representable as a float. To combat this we could explicitly do a cast using the CompCert float library before printing the floats?

Thoughts?

Zdancewic:
I see.  

It does seem like the better solution would be to disambiguate the floating-point syntactic constructs according to their intended bitwidths.  To do that, we'd probably want to add `EXP_Double` (and maybe even `EXP_Half`?).  The parser would then parse strings as it does currently, but then immediately try to use the ambient type information to convert literals into the right size, which could fail if the string isn't an exact representation of the floating-point value.  This is a bit painful, but doable, I think.

Chobbes:
@Zdancewic we could probably avoid doing this in the parser if we really want to, right?  We could do something similar to the recent `dtyp` pass over the AST, when we more easily have type information available? Or is this what you mean?

I guess one potential issue is that LLVM actually supports 128-bit floats, and maybe we should default to those in order to avoid precision loss?

I think we should definitely look into distinguishing this. I'm curious about the parser, as I haven't had a chance to play with menhir (and I've heard good things!), so if this is a route we want to take, I'd be happy to chip in :).

vzaliva:
Few comments to address various issues raised in the discussion above:

1. @Chobbes I've removed my remark about endianess because I checked and it is OK. I actually set up a VM with big-endian Debian and checked there :)

2. This patch fixes a problem where the LLVM IR program parsed from IR and pretty printed back might not compile due to float literals.

3. The problems where AST could be constructed with illegal float literals still remains. Additionally, there is a problem where parsed AST loses formal guarantees that 32-bit floats are indeed fit 32 bits and this information could not be used in program transformation proofs. I think further discussion of these in #95 is leading us towards the bit-length indexed representation of floating point literals in AST. If we go this way, this patch will become unnecessary.

Zdancewic:
OK, as an experiment I split the old AST constructor `EXP_Float` into `EXP_Float` and `EXP_Double` and did the plumbing to resolve the distinction in the parser.  This looks pretty reasonable to me (though if we want to also support half and 128-bit floats, we might want a more principled way of doing it).  

One thing I'm not 100% sure about is the implementation of `coqfloat32_of_string` (in `llvm_parser.mly`), which should fail if the string doesn't represent a valid 32-bit floating point number.  However, since OCaml doesn't have 32-bit floats it just uses OCaml's `float_of_string` function, which I don't believe is correct here.  We either need to call out to C or implement our own such routine.

Chobbes:
@Zdancewic something seems to be wrong with this for floats?

```
calvin@lappy src $ rm -rf output && make && ./vellvm --test-pp-dir ../tests/llvm-arith/float
make coq
make[1]: Entering directory '/home/calvin/git/vellvm/src'
make[1]: Nothing to be done for 'coq'.
make[1]: Leaving directory '/home/calvin/git/vellvm/src'
make extracted
make[1]: Entering directory '/home/calvin/git/vellvm/src'
make[1]: Nothing to be done for 'extracted'.
make[1]: Leaving directory '/home/calvin/git/vellvm/src'
make vellvm
make[1]: Entering directory '/home/calvin/git/vellvm/src'
Compiling Vellvm
dune build ml/main.exe
make[1]: Leaving directory '/home/calvin/git/vellvm/src'
(* -------- Vellvm Test Harness -------- *)
Running llc on: ../tests/llvm-arith/float/i8_uitofp_float.ll
Running llc on: ../tests/llvm-arith/float/hex_float_literal.ll
Running llc on: ../tests/llvm-arith/float/zero_double_literal.ll
Running llc on: ../tests/llvm-arith/float/float_literal.ll
Running llc on: ../tests/llvm-arith/float/zero_float_literal.ll
Running llc on: ../tests/llvm-arith/float/double_literal.ll

Parsing files in: ../tests/llvm-arith/float:
  OK
(*-------------------------------------- *)
Passed: 6/6
Failed: 0/6

calvin@lappy src $ cat output/double_literal.v.ll                                      [14:14]
define double @main(i8 %argc, i8** %arcv) {
; <label> 0
  
  
  ret double 0x405F547AE147AE14
}

calvin@lappy src $ cat output/float_literal.v.ll                                       [14:14]
```

No output for the `float_literal.ll` test file? I'm not immediately sure what the issue is.

Also these conversions remind me --- aren't Ocaml ints 31-bits because of the tagged pointer stuff? Will this cause us problems when trying to represent 32-bit floating point values? Are 64-bit ints tagged as well?

vzaliva:
Thanks! It looks reasonable after a quick glance.

One idea for dealing with `coqfloat32_of_string` problem: I think we can write in Coq (using Floq) a function which checks if given 64-bit binary float value could be exactly represented as 32 bit one. Then we can extract this function and use in the parser.

Zdancewic:
OK -- I'lm going to merge this into master now.  It's largely orthogonal to the semantic refactoring so it's good to get it in sooner anyway.

Issue #97, opened 2019-05-02
Check if floating point literal fits expected bit length


vzaliva:
In https://github.com/vellvm/vellvm/pull/96 @Zdancewic  wrote:

> One thing I'm not 100% sure about is the implementation of coqfloat32_of_string (in llvm_parser.mly), which should fail if the string doesn't represent a valid 32-bit floating point number. However, since OCaml doesn't have 32-bit floats it just uses OCaml's float_of_string function, which I don't believe is correct here. We either need to call out to C or implement our own such routine.

vzaliva:
My idea is instead of using C implementation write fitness check in Coq and extract it and use in the parser. I asked @zoickx for help and here is his response:

The `Binary.bounded` function from Flocq (http://kutt.ztld.org/bounded)
with a given `precision` and `emax` (the two parameters defining sets like b32 and b64)
can check if any given pair `m` `e` (mantissa, exponent) belongs to the set.

Normalization of that pair for a given `prec` `emax` in Flocq is performed via `shl_align_fexp` (http://kutt.ztld.org/shl_align_fexp/).
For inputs that can not be normalized, `shl_align_fexp` returns the same pair as the input. This, of course, creates a problem of discerning whether or not the input is already normal.
As far as I know, there are no wrappers around these functions doing what we want or even making them any nicer.

My take on it:
```Require Import ZArith.
Require Import Flocq.IEEE754.Binary.

Definition can_convert_exactly (target_prec target_emax : Z) (m : positive) (e : Z) : bool :=
  let '(res_m, res_e) := shl_align_fexp target_prec target_emax m e in
  bounded target_prec target_emax res_m res_e.

Definition can_convert_exactly_to_b32 := can_convert_exactly 24 128.
Definition can_convert_exactly_to_b64 := can_convert_exactly 53 1024.
    
Example one_quarter_is_b32 :
  can_convert_exactly_to_b32 1 (-2) = true.
Proof. reflexivity. Qed.
```

Notice that starting format does not matter to these functions at all. If you are asking specifically about converting something of "type b64", I would need a clarification on what you mean by "type b64"

vzaliva:
I did a first take on this. Now it fails `125.31999969482421875` and `1.0` literals.
Not sure what's wrong. I will investigate further.

vzaliva:
quick update. There is no existing function in Flocq to perform this check but we will try to define one.

vzaliva:
[...]

PR #99, opened 2019-05-21
Issue 97

vzaliva:
Fixes for https://github.com/vellvm/vellvm/issues/97

1. Implemented in Coq check if given 64-bit float could be converted to a 32-bit float without precision loss. Using extracted versions of these functions in parser to validate floating point literals.
2. Correctness proofs of functions used in the check.
3. Included more floating point unit tests in the test suite
4. Added new class of unit test which allow checking when the parser fails with a specific message.

( Contains contributions by @zoickx )


Zdancewic:
Thanks!  This looks good.  I'll merge it into master early next week (I'm currently chairing a workshop in Europe).

Issue #107, opened 2019-10-14
Show instance generating stack overflows


YaZko:
Trying to print big integers generates extracted code that fails with a stack overflow error.

This can be reproduced for instance by interpreting `../tests/llvm-arith/i32/sub_nsw.ll` on 0d12d9165997c3626be19052972a3bf3da733ea5 due to the debug event generating when entering `denote_exp`.

It seems to be somehow due to the `Show_Z` instance from `ext_lib`, but it's unclear to me at the moment.

Li-yao suggests burning the `Show` typeclass to the ground and use his s-expression library instead (https://github.com/Lysxia/coq-ceres).

YaZko:
PR #109 is on its way to address this issue.

PR #109, opened 2019-10-16
Removing Extlib's Show in favor for Ceres's Serialize

YaZko:
This pull request adresses #107 by removing altogether the use of Show.
It instead relies on the Ceres library that maintains a representation as S-expressions.
The pull request is _not_ yet ready to be merged as the conversion to string returns a representation of the S-expression, we need a small pretty-printer on top of it.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "lib/InteractionTrees"]
 	path = lib/InteractionTrees
 	url = https://github.com/DeepSpec/InteractionTrees.git
+[submodule "lib/coq-ceres"]
+	path = lib/coq-ceres
+	url = https://github.com/Lysxia/coq-ceres.git
--- a/src/Makefile
+++ b/src/Makefile
@@ -35,7 +35,11 @@ itrees:
 	git submodule update --init
 	make -C $(ITREEDIR)
 
-.PHONY: extracted itrees
+ceres:
+	git submodule update --init
+	make -C $(CERESDIR)
+
+.PHONY: extracted itrees ceres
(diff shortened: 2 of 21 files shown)


Lysxia:
What kind of pretty-printed output are you looking for? I see weird brackets (`[`, `<`) inside those `sexp`s... Do you want to print syntactically valid llvm? A proper pretty-printing library might be a better solution. The main selling point of sexps is precisely when you don't really care what the output looks like.

YaZko:
I think it's actually fine the way it is right now with the last few commits.
I do somewhat care about the look of the output, but only for the purpose of printing debug and error messages, not for printing valid syntax (we have an actual pretty printer for that).
The first output I got was really off putting, but partly because I was still introducing spaces manually that were redundant with the ones introduced between elements of the sexp by to_string, and partly because I was not collapsing enough elements into a single Raw as you explained this morning.
One thing that could still be useful I think would be to have a way to specify whether parentheses should be introduced or not for a given node of the sexp when converted to a string.

Lysxia:
I could imagine a separate pretty-printer that understands a few commands, so for example the S-exp

```
(%[, bar thon (%<; saumon carpe))
```

would get printed as

```
[bar, thon, <saumon; carpe>]
```

Or you could embrace S-expressions for what they are!

```
(array
  bar thon (spiky-array
            saumon carpe))
```

conveys the same information.

YaZko:
Sure, but getting debug information that looks fairly close to llvm syntax when you debug a llvm program is quite comfortable :)
But that is fair, I need to play a bit with it to really see what is or is not annoying.

PR #112, opened 2019-10-16
Experiment for better support for user intrinsics

YaZko:
At the moment, adding intrinsics requires to directly edit `IntrinsicDefinitions.v`, which is inconvenient.
We would like to have a way for this to take place outside of the code base.

This pull request parameterizes the model and the interpreter by a list of semantic-definitions, and defines the current instances by providing the empty list. As such, a user can define his own interpreter/model by providing additional intrinsics as illustrated in `UserTopLevelExample.v`.

Such a simple approach still requires to edit the front end however for it to work over the user's interpreter rather than the default one.

It also rises questions about results we will prove about the model: what should we assume about the user's intrinsics to prove the correctness of optimization parametrically in those?


src/coq/Handlers/Intrinsics.v
```diff
@@ -88,31 +88,31 @@ Module Make(A:MemoryAddress.ADDRESS)(LLVMIO: LLVM_INTERACTIONS(A)).
...
-     SAZ: This definition is trickier than one wants it to be because of the 
-     dependent pattern matching.  The index of the IntrinsicE constructor need to 
+     SAZ: This definition is trickier than one wants it to be because of the
+     dependent pattern matching.  The index of the IntrinsicE constructor need to
      be used to coerce the result back to the general ITree type.
 
-     We solve it by using the "Convoy Pattern" (see Chlipala's CPDT).  
+     We solve it by using the "Convoy Pattern" (see Chlipala's CPDT).
 
      IntrinsicE ~> LLVM _MCFG1
```
Zdancewic: This comment is out of date.

vzaliva:
I've tested it and it works well. I was able to plug my own definitions using this mechanism.
So, in general, I would love to see this merged ASAP. Thanks!
One minor remark: Perhaps `src/coq/UserTopLevelExample.v` should be moved elsewhere, as it is not the part of Vellvm core implementation.

YaZko:
Great, I'll move away and comment `UserTopLevelExample` and will merge this this afternoon then.
Thanks!

YaZko:
I simply removed the file and merged the PR into master.

PR #120, opened 2019-12-05
Make functions take uvalues instead of dvalues.

Chobbes:
This addresses #119.

Making a pull request because this ended up being a fairly substantial change, and I think it would be good to have a few eyes on it before we merge, in case there are any concerns.

Most of it is fairly straightforward. I made intrinsics take `uvalue`s as well, because it's reasonable that you might want to model nondeterminism there as well. This resulted in a pick before `memcpy`, which currently has the proposition that all of the arguments to `memcpy` have a unique concretization (or it's `UB`).

Intrinsics for floating point operations currently just fail if the value isn't concrete. Should we pick here instead? Now that I think about it, this is probably a more reasonable first step.

And the main thing I have questions about is `Renaming.v`. I pushed some changes that I made to try to get some of the proofs to work, but I'm not super familiar with the stuff in that section, so I ended up commenting it out. Does anybody know much about that?

Thanks!

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,6 +4,3 @@
 [submodule "lib/InteractionTrees"]
 	path = lib/InteractionTrees
 	url = https://github.com/DeepSpec/InteractionTrees.git
-[submodule "lib/coq-ceres"]
-	path = lib/coq-ceres
-	url = https://github.com/Lysxia/coq-ceres.git
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,7 @@ install:
   - opam install -y --verbose -j ${NJOBS} coq-ext-lib
   - opam install -y --verbose -j ${NJOBS} coq-flocq
   - opam install -y --verbose -j ${NJOBS} coq-paco
+  - opam install -y --verbose -j ${NJOBS} coq-ceres
   # This branch uses the lib/InteractionTree submodule.
   # - opam pin -y add coq-itree --dev-repo
   # - opam install -y --verbose -j ${NJOBS} coq-itree
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ University of Pennsylvania as part of the DeepSpec project.
     - flocq      (installed via, e.g. opam install coq-flocq, see note below) 
     - itree      ~~(installed via, e.g. opam install coq-itree)~~
       - Currently you should actually just use the submodule (lib/InteractionTrees)
-    - ceres      (installed via the submodule (lib/coq-ceres))
+    - ceres      (installed via, e.g. opam install coq-ceres)
 - ocamlc : version 4.04    (probably works with 4.03 or later)
   - OPAM packages: dune, menhir, [optional: llvm  (for llvm v. 3.8)]
 
--- a/src/Makefile
+++ b/src/Makefile
@@ -7,9 +7,8 @@ MLDIR = ml
 EXTRACTDIR = ml/extracted
 
 ITREEDIR=../lib/InteractionTrees
-CERESDIR=../lib/coq-ceres
 
-COQINCLUDES=$(foreach d, $(COQDIR), -R $(d) Vellvm) -R $(EXTRACTDIR) Extract -R $(ITREEDIR)/theories/ ITree -R $(CERESDIR)/theories/ Ceres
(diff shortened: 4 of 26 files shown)


YaZko:
I am not sure that we want to require the intrinsics to be defined over `uvalues`, the type `semantic_function` should remain the same I think. 

They morally are defined outside of LLVM, in particular in our case exposed at the interface in Gallina, and as such should not have to be concerned with this notion of `uvalue` that is rather internal.

In particular btw that's what you did with `mem_cpy`, you left it over `dvalue`, and picked at the interface. I would have done the sane in denotation for non-memory dependent intrinsics.

Chobbes:
I'm not really sure, I could see this going either way?

I think you're right that intrinsic functions are kind of external to LLVM and thus shouldn't really have a concept of uvalues, but I also wonder if you want to pass uvalues anyway and pass the responsibility of concretizing these onto the intrinsic handlers --- after all they can always pick as well.

In particular with something like `memcpy` LangRef says things like:

> The ‘llvm.memcpy.*’ intrinsics copy a block of memory from the source location to the destination location, which are not allowed to overlap.

Which suggests to me that you might have `UB` with intrinsics? And, I assume then that, if the destination of `memcpy` isn't unique this should be `UB` just like how it's `UB` if the destination of a store instruction isn't unique... Let's assume this is the case because it's a bit easier to talk about than the above.

So, I might want to raise UB if `memcpy` gets a destination which has multiple concretizations, but I don't want to enforce the unique concretization constraint on `uvalue`s for every intrinsic --- just the ones like `memcpy` where this (presumably) results in UB. If intrinsic events are constructed with `dvalue`s then I need to pick before I build the intrinsic event...

```coq
        (* Call *)
        | (pt, INSTR_Call (dt, f) args) =>
          uvs <- map_monad (fun '(t, op) => (translate exp_E_to_instr_E (denote_exp (Some t) op))) args ;;
          returned_value <-
          match Intrinsics.intrinsic_exp f with
          | Some s => trigger (Intrinsic dt s uvs)
          | None =>
            fv <- translate exp_E_to_instr_E (denote_exp None f) ;;
            (* debug ("Call to function: " ++ to_string f) ;; *)
            trigger (Call dt fv uvs)
          end
          ;;
          match pt with
          | IVoid _ => ret tt
          | IId id  => trigger (LocalWrite id returned_value)
          end
```

So, I guess when denoting the call I would need to figure out which kind of intrinsic I have and call `pick` with the appropriate predicate before triggering the intrinsic event... Or maybe the list of intrinsics could also have a list of "pick" functions, or something.

I am unsure if this is the kind of thing we need to handle, but it seems conceptually nice to have the intrinsic itself decide if being called nondeterministically matters as opposed to the above?

Again, this is maybe all a moot point, because maybe it doesn't make sense to talk about UB with respect to intrinsics, but if it does make sense I think we might also want intrinsics to take `uvalue`s? I'm unclear on this.

Zdancewic:
I suspect that the LLVM developers would consider some intrinsics (like memcpy) to be possible sources of undefined behavior.  At the moment, our semantics requires intrinsics to be pure functions from dvalues to dvalue.  If we wanted to allow the intrinsics to do Pick or raise UB, we'd have to make the denotation of an intrinsic be an itree-returning function, which might complicate the Intrinsics modules dependencies a little bit, but should be doable.

In the long run, we probably do want to have non-pure intrinsics, though, so I think we could move to uvalues in the interface.

Zdancewic:
On the other hand, I'm sympathetic to Yannick's point that things outside of the Vellvm semantics shouldn't necessarily know about uvalue's structure.  Maybe we should think about making some kind of abstraction to hide those details from the outside?  i.e. our current uvalue datatype could be the implementation of a structure satisfying the interface, which would probably need a way to inject "primitive" values and offer the Pick event...

Chobbes:
Okay. So, to summarize from the meeting for future reference it sounds like the plan is:

- Keep intrinsics as functions from lists of dvalues to dvalues
- Add an external call event which is from lists of dvalues to dvalues
- Introduce picks which either have the predicate `True` or the unique concretization predicate
- Possibly in the future we will extend intrinsic / external call events with predicates like pick

Let me know if this seems off, or like I'm forgetting something important. In the meantime, I'm going to make these changes and update the PR :).

Chobbes:
I think this might be good to go now, if you want to have a look at it :).

PR #124, opened 2020-02-11
do not git update itress

vzaliva:
automatic submodule update prevents from working with older checked out version on vellvm. 
it updates itrees automatically during build and new version of itrees might be incompatible with checked out version of  vellcm.


Zdancewic:
Yes, which version of the ITrees library to link to is a bit of an issue.  I like the idea of making the update manual.

Issue #125, opened 2020-02-14
Denotation of open cfgs


YaZko:
The function denote_cfg allows us to denote _some_ notion of open cfg: it takes a list of blocks and denote them in such a way that the denotation of a full cfg can be defined by instantiating it over the list of all blocks in the cfg, and its entry point as argument.
It is however not a properly compositional way to do so. Abstracting away a little bit, the function is the following:

         [|bks|] = loop (fun bid =>
                     if bid ∈ (ids bks)                       (* over ids of blocks considered, we evaluate the block *)
                     then bov <- [|bks[bid]|]
                          match bov with
                          | inr v    => ret (inr (inr v))     (* if it returns, we terminate *)
                          | inl bid' => ret (inl bid')        (* otherwise we _always_ feed it back to the loop *)
                     else               ret (inr (inl bid))   (* over ids of blocks not considered, it's the id, we return it immediately *)
                     )

Intuitively, this means that we tie the recursive knot over all available blocks greedily. In order to be able to define combinators, we need to be a bit more subtle, in the style of [Asm].
The branch `open-cfg` plans to do such an extension. The current plan is the following:

Let (Domain := list block_id) and bks be a list of blocks.
We want the denotation to take two domain parameters as arguments: ([|bks|] A B) corresponds
to the denotation of bks where one can only enter the subcfg through one of the labels provided
in A, and do not loop back from a label described in B (or from the opposite perspective,
B are the possibly returned labels).
Define dom(bks) := {l | l ∈ map blk_id bks}
Define codom(bks) := {l | l ∈ map blk_term bks}
Compared to Asm, we try a slightly less typed approach: instead of defining the domain of the denotation function itself as A, we have it as a total function over label ids, and maintain invariants instead.
Assume the following invariants:
         A ⊆ dom(bks)
         B ⊆ codom(bks)
         dom(bks) \ A ~~ codom(bks) \ B
Define I := cod(bks) \ B
Then it should be correct to redefine [|bks|] as follows:

         [|bks|] A B = loop (fun bid =>
                       if bid ∈ A                                 (* over ids in A, we evaluate the block *)
                       then bov <- [|bks[bid]|]
                          match bov with
                          | inr v    => ret (inr (inr v))         (* if it returns, we terminate *)
                          | inl bid' => if bid' ∈ B
                                        then ret (inr (inr bid')) (* if we get something in B, we return *)
                                        else ret (inl bid')       (* otherwise it's in I, hence we loop *)
[...]

PR #126, opened 2020-02-16
Open cfg

YaZko:
This ongoing PR attempts to address #125. 

It currently introduces  a data-type of open-cfg that explicitly carries an input and output interface that controls in particular when the denotation should re-entry the loop and when it should return the computed label.

In contrast to the [asm] language from the itree tutorial, the returned type of the denotation is less precise: it is not explicit that it can only return a label that belongs to the output interface. This will have to be compensated with invariants.

Symmetrically, the denotation can be called on any label. It is slightly unclear whether it should fail or behave as the identity over labels that are not in the input interface. It might not matter.

It is at the moment slightly unclear what should be done with respect to phi-nodes: the current scheme denoting the future ones at the end of the body of the loop is invalid.

The next steps are to figure out which combinators over [open_cfg] we want and prove them correct.

### DIFF
--- a/src/coq/Denotation.v
+++ b/src/coq/Denotation.v
@@ -873,7 +873,67 @@ Module Denotation(A:MemoryAddress.ADDRESS)(LLVMEvents:LLVM_INTERACTIONS(A)).
 
        *)
 
-      Definition denote_bks (bks: list _): block_id -> itree instr_E (block_id + uvalue) :=
+      Definition interface: Type := list block_id.
+      Definition open_cfg: Type  := list (block dtyp).
+      Definition dom: open_cfg -> interface := List.map blk_id.
+      Definition lbl_from_terminator {T} (t: terminator T): interface :=
+        match t with
+        | TERM_Br _ id1 id2 => [id1;id2]
+        | TERM_Br_1 id => [id]
+        | _ => []
+        end.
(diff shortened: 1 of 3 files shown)


YaZko:
Note: to get proper combinators, we need a coproduct structure on the denotation. This requires (?) the signature of the body to be more symmetric, i.e. be `(bid + uvalue) -> itree E (bid + uvalue)` rather than the current more sensible `bid -> itree E (bid + uvalue)`. It's a bit weird.

YaZko:
Note: I've started trying to commit to strongly typed denotation by working over the the subcategory of ktrees over the following finite types:

`{ x : (block_id * nat) | nth_error l (snd x) = Some (fst x) }`

It admits unicity of proofs defining the subset so it's fairly easy to work with. (See `FinLabels.v`).

My new concern is that contrary to `asm`, the language is not purely side-effect, a block can return a value. Does this mean I should work over the quite ugly:

`{ x : ((block_id * nat) + uvalue) | (exists y, x = inl y) -> nth_error l (snd x) = Some (fst x) }`?

It feels wrong, but I'm not sure I see a way out for now.

YaZko:
This has stalled for too long, I'm closing it. I'll revisit it some day if further experiment suggest it might be worth it.

PR #128, opened 2020-05-14
Moving the potential failure in conversion from tle to mcfg down to the parser?

YaZko:
@Zdancewic This does not address at all the interesting semantics questions that we were discussing this afternoon, but do you thing that such a change would be sensible? 

This patch is partial, it needs some OCaml code. But basically it is simply enforcing the parser to output pairs of a block and a list of block, the former being the initial block of the cfg, so that on the Coq side the conversion between toplevel entities and mcfg can be total.

The change is cosmetic, but would allow for cleaner lemmas to manipulate the toplevel semantics I think.

### DIFF
--- a/src/coq/CFG.v
+++ b/src/coq/CFG.v
@@ -189,30 +189,25 @@ Definition phis_of_definition (d:definition (list block)) block_id : option bloc
   block_to_phis b.
 *)
 
-Definition init_of_definition (d: definition T (list (block T))) : option block_id :=
-  match (df_instrs d) with
-  | [] => None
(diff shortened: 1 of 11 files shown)


Zdancewic:
This is a great idea.  I've made the necessary changes to the parser/pretty printer infrastructure.  Everything should build in this branch now -- if you agree, let's go ahead and merge this into master.

YaZko:
Oh I didn't mean to push the coding the OCaml bit burden on you :) But this is great, thanks a lot! I'll merge this in.

PR #134, opened 2020-07-22
Issue 132

vzaliva:
Fix for https://github.com/vellvm/vellvm/tree/issue-132
Signaling/Quietness of NaNs is preserved.


Zdancewic:
Thanks!

PR #143, opened 2020-10-23
Testing vvijay

Vighnesh-V:
This PR adds support for an inline poison assertion of the form:
```
; ASSERT POISON: call <TYPE> @<fname>(...)
```
as well as migrating the test cases in `test.ml` over to the new inline assertion test format.


YaZko (approved):
This looks like some excellent job @Vighnesh-V, thanks!
I'll let Steve double check before merging as he's a bit more familiar than I am with the testing infrastructure, but I don't see any issue at first glance.
We can now try to import other test suites from the literature :)

Zdancewic (approved):
This looks awesome!  Thanks.

PR #145, opened 2020-11-23
Restructuring Memory module, no duplicate LLVMEvent

euisuny:
Restructured so that : 
1. `Memory.v` is split into `Memory.v` (memory representation) and `MemoryTheory.v` (theory of memory representation)
2. We can still parameterize on a memory model without having to deal with duplicate representations of `LLVMEvent`, as is the current issue on `master` now.


YaZko:
Thanks a lot @euisuny !

Just a nitpick: currently in master we don't have a duplication of LLVMEvent (that would prevent compilation), but to avoid it we inherit the LLVMEvent implementation from the MemoryTheory module, such that the other handlers depend on the memory where they shouldn't.

I am really confused by the use of a module type here, but well done at any rate!!

PR #147, opened 2021-02-12
Simplification of the intrinsics

YaZko:
The semantics currently takes as parameter a field `user_intrinsics` that I believe is only crust from iterations on the design.

The initial intention was as follow:
- VIR provides a set of standard intrinsics whose names are collected in `defined_intrinsics`
- users can define their own intrinsics whose name are collected in `user_intrinsics`
- the top-level semantics takes `user_intrinsics` as argument.

However:
- these user intrinsics cannot be linked outside of the development, it needs to be recompiled anyway
- we have never used these user intrinsics
- it shows: there were a minor bug relating to them
- they are annoying as everything everywhere needs to be quantified by this useless parameter of `user_intrinsics`.

So I just removed completely this useless parameter in this branch.

Before merging I wanted to double check that we agree on this direction?

### DIFF
--- a/src/coq/Handlers/Intrinsics.v
+++ b/src/coq/Handlers/Intrinsics.v
@@ -77,20 +77,19 @@ Module Make(A:MemoryAddress.ADDRESS)(LLVMIO: LLVM_INTERACTIONS(A)).
      semantic functions.
    *)
 
-  Definition defs_assoc (user_intrinsics: intrinsic_definitions) := List.map (fun '(a,b) =>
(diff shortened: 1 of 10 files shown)


YaZko:
I should specify that I have recompiled Helix and run the tests against this branch.

Chobbes:
Looks good to me! I was initially a little worried because I vaguely remembered that helix used some extra intrinsics, but after a conversation on slack it sounds like those got incorporated into Vellvm proper.

PR #148, opened 2021-02-14
support for c string literals

Zdancewic:
Added support for LLVM IR's c string literals.  They are translated to/from i8 arrays during parsing/pretty printing. 

Submitting as a pull request to avoid any unintentional interference with HELIX.


YaZko:
Awesome, that'll help a lot for testing!

I tested it against HELIX, nothing breaks, so I just merged it into master.

Issue #115, opened 2019-10-20
Can't parse "nsw nuw"!


Chobbes:
LLVM instructions like `shl` can have both of these flags like so:

`%b = shl nsw nuw i32 undef, %a`

However, the parser only handles them in this order:

`%b = shl nuw nsw i32 undef, %a`

But, LLVM itself is fine with either order.

Zdancewic:
Fixed in #157.

PR #157, opened 2021-03-11
Anon ids

Zdancewic:
A bunch of cleanup in the front end.  The main changes are:

1. Fix handling of anonymous / implicit identifiers for labels, etc.
2. Extend syntax with LLVM 11 function attributes
3. Make switch indices literal ints rather than arbitrary expressions
4. Permit nuw and nsw to permute (issue #115) 
5. Add some more test cases for these features

I think this will also address @YaZko's issue about branching to "numbered" blocks.  See `tests/ll/anon_label.ll` for an example test case.

As a plus, we can now parse code generated by Clang or llc for Oat's runtime.c (which I haven't added to the repo yet).


YaZko:
This is looking great! Well except for the daunting diff on `fn_attr`, but that's part of the job :-)

I just tested it locally, it solves my problem with anonymous blocks definitely, thanks! 

I would be curious to see in more details the story for the runtime you mention. 
Notably because I have been trying to test the switch semantics and others by compiling some examples from llvm's test suite, which is provided in term of C programs rather than IR ones, but it generates stuff like:
```
declare i32 @printf(i8*, ...) #1
(...)
  %30 = call i32 (i8*, ...) @printf(i8* getelementptr inbounds ([7 x i8], [7 x i8]* @.str, i64 0, i64 0))
```
I am not sure of the semantics of the ellipsis, but we fail to parse it currently.

Zdancewic:
The `...` is llvm's notation for a variable argument function prototype, which is needed to call/declare `printf`.  I started to add vararg support (see the "unused token DOTDOTDOT" message when compiling), but the semantics will be a bit tricky and even LangRef doesn't really specify where such varargs are valid.  (e.g. can you have more than one?).

That said, support for varargs and maybe some form of printf would be very useful for testing, so it's on my short list.

YaZko:
I see, makes sense. Thanks!

Zdancewic:
I don't know why travis-ci fails.  This builds just fine for me.  (And I guess for others.). If it that seems OK, perhaps we can go ahead and merge.

YaZko:
I built it locally without issue, we can definitely go ahead and merge!

I believe that Travis is failing because when QuickChick got added to the compilation process, it introduced a new dependency on `ocamlbuild` that has not been fed to Travis.

PR #162, opened 2021-06-03
Print the error message passed to raise as a side effect instead of including it in the failure event

nchappe:
This introduces a `print_msg` function that is a no-op in Coq but extracted to an actual printing function in OCaml.
This allows to prove properties like `eutt eq (raise msg1) (raise msg2)`.

### DIFF
--- a/src/coq/Semantics/LLVMEvents.v
+++ b/src/coq/Semantics/LLVMEvents.v
@@ -95,10 +95,14 @@ Set Contextual Implicit.
   Definition debug {E} `{DebugE -< E} (msg : string) : itree E unit :=
     trigger (Debug msg).
 
-  Definition FailureE := exceptE string.
+  Definition FailureE := exceptE unit.
+
+  (* This function can be replaced with print_string during extraction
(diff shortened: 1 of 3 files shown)


YaZko:
As Nicolas says, the issue is that we are bound to enforce syntactic equality over the failure events when proving a transformation. However, for debugging purpose, we carry syntax in our failure events, which makes proving transformations that perform renaming for instance impossible for no good reason.

In particular, I had the same issue in the block fusion proof, but lazily resorted on removing the annoying bit from the error message that was blocking me.
A possible alternative would be to reason after interpretation of the failure event into a failure monad, or alternatively have a translation into a non-informative failure event, but both are quite heavy weight.

This idea that Nicolas had is pleasantly lightway: basically the string is discarded on the Coq side, but recovered by extraction. We of course want to be careful not to mess up through extraction with the semantics too much as it would quickly become dubious, but this specific case seems fair to me.

Zdancewic:
This is a great idea!  Thanks.

PR #191, opened 2022-02-09
Add modul_of_mcfg

atrieu:
I've been playing a bit around with VeLLVM for writing LLVM IR transformations and trying to prove them correct. It seemed like the most logical way was to write a transformation as a `mcfg -> mcfg` function in Coq, though I didn't find a way to print back the resulting `mcfg` into a `.ll` file. Doing the transformation at the `modul` level and then reasoning at the `mcfg` level for the semantics seemed strange (to me, at least).

This PR adds a `modul_of_mcfg` function that can transform back a `mcfg` into a `modul (block T * list (block T))` that can be properly consumed by `Llvm_printer.modul` for printing. There is also an additional lemma to show that `modul_of_mcfg` and `mcfg_of_modul` are inverse of each other.

There is also a `toplevel_entities_of_modul` to transform back a `modul` into a `toplevel_entities`, though with no associated proof. I thought this was necessary because `Driver.ml` only used `Llvm_printer.toplevel_entities` and I didn't see `Llvm_printer.modul` until I started preparing this PR...

### DIFF
--- a/src/coq/Syntax/CFG.v
+++ b/src/coq/Syntax/CFG.v
@@ -156,6 +156,58 @@ Section CFG.
       m_definitions := flat_map definitions_of tles;
     |}.
 
+  Definition tle_of_name {X} (m_name: option string): toplevel_entities T X :=
+    match m_name with
+    | None => []
+    | Some n => [TLE_Source_filename n]
+    end.
+
+  Definition tle_of_target {X} (m_target: option string): toplevel_entities T X :=
+    match m_target with
+    | None => []
+    | Some n => [TLE_Target n]
+    end.
+
+  Definition tle_of_datalayout {X} (m_datalayout: option string): toplevel_entities T X :=
+    match m_datalayout with
+    | None => []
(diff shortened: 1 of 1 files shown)


Chobbes:
Ah, nice! This looks good to me.

Also a heads up for the potentially distant future --- I'm hoping to have `coq/QC/ShowAST.v` become the main way that we print LLVM files (I'm hoping we slowly move more of the ml stuff into Coq, anyway). Currently it's a little incomplete and you should stick to `Llvm_printer.modul`, and it may be a long time before we actually switch. We use `coq/QC/ShowAST.v` to print out LLVM files for our QC tests because it's a pain to get the `.ml` files included for QC.

YaZko:
Ah I messed up: we had both a master and a dev branches around, and I just killed the former to clean up things. However this PR was made against `master` and hence got closed in the process.

@atrieu Would you mind reopening it against `dev` so that I can merge it? Apologies for the annoyance!

PR #192, opened 2022-03-02
Fix itree coq library name in readme

saulshanabrook:
I believe the library has no `s` on the end.


YaZko:
Indeed, thanks for catching that!

PR #196, opened 2022-03-07
Document that coq quickcheck library is required

saulshanabrook:
Mentions in the README and in the makefile that [coq-quickcheck](https://github.com/QuickChick/QuickChick) is required now as well to build the package.


Chobbes:
We're actually using a custom version of `QuickChick` that's in `lib/QuickChick`. The regular `QuickChick` library won't work with the generators for `Vellvm`.

This library should be built automatically with `make`, though it is a submodule so you have to make sure you have it. I believe running:

```
git submodule update --init --recursive
```

Should grab it if you didn't `git clone --recursive`.

saulshanabrook:
Ah ok thanks for the pointer!

Chobbes:
No problem! Let me know if there are any problems, though. I also added a note in the install instructions about this.

PR #220, opened 2022-05-31
Memory testing gen solved aggregate and vector element bug

hanxic:
Edit to both let vector to have size greater than 0 and generate expression such that the element of aggregate types and vector type to not have local identifier

### DIFF
--- a/src/coq/QC/GenAST.v
+++ b/src/coq/QC/GenAST.v
@@ -378,19 +378,23 @@ Section TypGenerators.
     | O => gen_sized_typ_0
     | (S sz') =>
       oneOf_LLVM
-        [ gen_sized_typ_0
+        ([ gen_sized_typ_0
         ; ret TYPE_Pointer <*> gen_sized_typ_size sz'
         (* TODO: Might want to restrict the size to something reasonable *)
         ; ret TYPE_Array <*> lift_GenLLVM genN <*> gen_sized_typ_size sz'
-        ; ret TYPE_Vector <*> lift_GenLLVM genN <*> gen_sized_typ_size sz'
+        ; ret TYPE_Vector <*> (n <- lift_GenLLVM genN;; ret (n + 1)%N) <*> gen_sized_typ_size 0
         (* TODO: I don't think functions are sized types? *)
         (* ; let n := Nat.div sz 2 in *)
         (*   ret TYPE_Function <*> gen_sized_typ_size n <*> listOf_LLVM (gen_sized_typ_size n) *)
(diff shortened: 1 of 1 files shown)


src/coq/QC/GenAST.v
```diff
@@ -957,6 +961,13 @@ Definition genType: G (typ) :=
 
   (* TODO: should make it much more likely to pick an identifier for
            better test cases *)
+
+  Fixpoint gen_exp_size_wo_ctx (sz: nat) (t : typ) {struct t} : GenLLVM (exp typ) :=
```
Chobbes: Is this function used anywhere? The body also seems suspicious...

This actually isn't a bad idea, though, we could make a function that temporarily clears the LLVM context and runs a generator in that. We should probably do that...

Something like:

```
clear_ctx {A} (g : GenLLVM A) : GenLLVM A
```

Which just runs `g` as though there's no context. Could be a little dangerous for generators that add stuff to the context, though. You'd have to make sure you actually get fresh variables (should be doable) and that any new ids generated get merged in (also doable).

hanxic: This function is not used. It is not used because it is redundant with gen_exp_size, which I simply change gen_size_0 and have been able to successfully generate types

hanxic: the reason why I’m not taking that approach is that it will get the probability mess up. Like if we generate a non-local identifier expression in clear_ctx and then add ctx to it in gen_exp_size, the probability for an expression in clear_ctx get picked is going to be lower than the one in the ctx

Chobbes (commented):
This looks good to me, but I think you have an extra function `gen_exp_size_wo_ctx` that we should probably remove (or use if it's useful somewhere, but it looks like the body is mostly a placeholder?)

hanxic:
Done deleting

src/coq/QC/GenAST.v
Chobbes: Okay, I think we should remove this function then. Push the change to remove it and I'll approve the PR and get it merged :).

You're right that `clear_ctx` would change the probability of generating some stuff, but so does setting `gen_size_0`, right? The idea is just that `clear_ctx` could be a general combinator that we could use to change the behaviour of certain generators. We don't have to build it now, though, but something to keep in mind!

The idea would be similar to how we use `replace_ctx` in a few places to hide variables that won't be visible in certain places, but a bit fancier.

Chobbes:
Looks great, thanks! :)

PR #225, opened 2022-06-02
Memory testing gen fix insert/ extract generator function

hanxic:
Run `make qc-tests` now can generate insert / extract element / value function

### DIFF
--- a/src/coq/QC/GenAST.v
+++ b/src/coq/QC/GenAST.v
@@ -286,7 +286,12 @@ Section GenerationState.
        let new_ctx := vars ++ ctx in
        modify (replace_ctx new_ctx);;
        ret tt.
-
+  Definition hide_ctx {A} (g: GenLLVM A) : GenLLVM A
+    := ctx <- get_ctx;;
+       modify (replace_ctx []);;
+       a <- g;;
+       append_to_ctx ctx;;
+       ret a.
   Definition append_to_typ_ctx (aliases : list (ident * typ)) : GenLLVM unit
     := ctx <- get_typ_ctx;;
        let new_ctx := aliases ++ ctx in
@@ -373,7 +378,7 @@ Section TypGenerators.
                 (* ; TYPE_Opaque *)
                 ])).
 
-  Program Fixpoint gen_sized_typ_size (sz : nat){measure sz} : GenLLVM typ :=
+  Program Fixpoint gen_sized_typ_size (sz : nat) {measure sz} : GenLLVM typ :=
     match sz with
     | O => gen_sized_typ_0
     | (S sz') =>
@@ -886,7 +891,9 @@ Definition get_ctx_agg_typ : GenLLVM (ident * typ) :=
   oneOf_LLVM (map ret aggs_in_context).
 
 Definition filter_vec_typs (ctx: list (ident * typ)) : list (ident * typ) :=
-
+  (* The function calls filter, which get all the vector type in the context
+   It does not utilize filter_type because we don't know the inner type of the vector
+   We simply wants to find all vectors (assume valid) in the context*)
   filter (fun '(_, t) =>
             match t with
             | TYPE_Vector _ _ => true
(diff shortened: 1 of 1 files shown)


src/coq/QC/GenAST.v
```diff
@@ -373,23 +373,22 @@ Section TypGenerators.
                 (* ; TYPE_Opaque *)
                 ])).
 
-  Program Fixpoint gen_sized_typ_size (sz : nat) {measure sz} : GenLLVM typ :=
+  Program Fixpoint gen_sized_typ_size (sz : nat){measure sz} : GenLLVM typ :=
```
Chobbes: Can you get rid of these whitespace changes to this function?

src/coq/QC/GenAST.v
```diff
@@ -940,39 +958,25 @@ Definition gen_insertelement : GenLLVM (typ * instr typ) :=
...
-  end with
-  gen_typ_eq_struct (fields: list typ) (gFields: GenLLVM (list (typ * exp typ))) {struct fields} : GenLLVM (exp typ):=
-  match fields with
-  | nil => l <- gFields;;
-  ret (EXP_Array l)
-  | hd::tl => l <- gFields;;
-  ex <- gen_typ_eq hd;;
-  gen_typ_eq_struct tl (ret ((hd, ex)::l))
-  end.
+(* TODO: code for ptrtoint incomplete
```
Chobbes: Let's take this out of this change for now.

src/coq/QC/GenAST.v
```diff
@@ -982,7 +986,48 @@ Definition genType: G (typ) :=
 
   (* TODO: should make it much more likely to pick an identifier for
            better test cases *)
+Fixpoint clear_ctx (t: typ) :=
```
Chobbes: What does this function actually do? I'd expect the `clear_ctx` that we talked about before to only take in an arbitrary generator (not necessarily one for expressions) and produce a new one.

src/coq/QC/GenAST.v
```diff
@@ -869,9 +886,10 @@ Definition get_ctx_agg_typ : GenLLVM (ident * typ) :=
   oneOf_LLVM (map ret aggs_in_context).
 
 Definition filter_vec_typs (ctx: list (ident * typ)) : list (ident * typ) :=
+
   filter (fun '(_, t) =>
```
Chobbes, line 896: Might be worth adding a little comment about what this filter function is for, I think it's not immediately obvious.

Chobbes: Specifically, what does this call to `filter` do?

src/coq/QC/GenAST.v
hanxic: I think I've adjust this. Please let me know if I miss something. Apologize in advance if I did

src/coq/QC/GenAST.v
hanxic: I've removed them from the code

src/coq/QC/GenAST.v
hanxic: Made hide_ctx. Deleted clear_ctx

src/coq/QC/GenAST.v
hanxic: filter will just filter all the vector types, assuming that they are valid. I don't think I can call filter_type because that one actually compares what's inside the vector and I don't want it. I just want to get all the vectors in the ctx.

src/coq/QC/GenAST.v
```diff
@@ -286,7 +286,12 @@ Section GenerationState.
        let new_ctx := vars ++ ctx in
        modify (replace_ctx new_ctx);;
        ret tt.
-
+  Definition hide_ctx {A} (g: GenLLVM A) : GenLLVM A
```
Chobbes, line 290: Let's add a newline between this definition and the previous one.

src/coq/QC/GenAST.v
```diff
@@ -286,7 +286,12 @@ Section GenerationState.
...
+    := ctx <- get_ctx;;
+       modify (replace_ctx []);;
+       a <- g;;
+       append_to_ctx ctx;;
+       ret a.
   Definition append_to_typ_ctx (aliases : list (ident * typ)) : GenLLVM unit
```
Chobbes, line 297: A newline between this definition and previous one would help with reading too.

src/coq/QC/GenAST.v
```diff
@@ -381,15 +386,14 @@ Section TypGenerators.
         ([ gen_sized_typ_0
         ; ret TYPE_Pointer <*> gen_sized_typ_size sz'
         (* TODO: Might want to restrict the size to something reasonable *)
-        ; ret TYPE_Array <*> lift_GenLLVM genN <*> gen_sized_typ_size sz'
+        ;ret TYPE_Array <*> lift_GenLLVM genN <*> gen_sized_typ_size sz'
```
Chobbes: Let's add this space back.

src/coq/QC/GenAST.v
```diff
@@ -823,15 +827,33 @@ Fixpoint get_index_paths_aux (t_from : typ) (pre_path : list Z) {struct t_from}:
...
+  | t => [(t, pre_path)]
+  end with
+  get_index_paths_from_struct_xinvec (fields: list typ) (pre_path: list Z) (current_index : Z) {struct fields}: list (typ * list Z) :=
+  match fields with
+  | nil => nil
+  | h::t => let head_list := map (fun '(t, p) => (t, pre_path ++ [current_index] ++ p)) (get_index_paths_aux_xinvec h []) in
+  let tail_list := get_index_paths_from_struct_xinvec t pre_path (current_index + 1%Z) in
+  head_list ++ tail_list
+  end.
+(*Example test1 : get_index_paths_aux_xinvec (TYPE_Struct [TYPE_Struct [TYPE_Struct [TYPE_I 8; TYPE_I 8]; TYPE_Vector 2 (TYPE_I 32)]; TYPE_Vector 3 (TYPE_I 1)]) [] = [].
```
Chobbes: Delete

src/coq/QC/GenAST.v
```diff
@@ -823,15 +827,33 @@ Fixpoint get_index_paths_aux (t_from : typ) (pre_path : list Z) {struct t_from}:
...
+Proof. simpl. *)
```
Chobbes: Delete

src/coq/QC/GenAST.v
```diff
@@ -823,15 +827,33 @@ Fixpoint get_index_paths_aux (t_from : typ) (pre_path : list Z) {struct t_from}:
...
 Definition get_index_paths_agg (t_from: typ) : list (typ * list (Z)) :=
```
Chobbes, line 852: Make sure to leave a newline between the previous definition and this one.

src/coq/QC/GenAST.v
Chobbes: Oh sorry, this isn't the filter I was confused about. Will add another comment...

src/coq/QC/GenAST.v
```diff
@@ -823,15 +827,33 @@ Fixpoint get_index_paths_aux (t_from : typ) (pre_path : list Z) {struct t_from}:
...
-  let edited_path := map (fun '(t, path) => match path with
-                                         | hd::tl => (t, tl)
-                                         | _ => (t, path)
-                                         end) (get_index_paths_aux t_from nil) in
+  let agg_paths := get_index_paths_aux_xinvec t_from nil in
   filter (fun '(_,path) => match path with
```
Chobbes: What does this filter do? We should explain this one. Sorry, wrong one last time.

src/coq/QC/GenAST.v
```diff
@@ -1209,9 +1205,9 @@ Section InstrGenerators.
            ret (TYPE_Pointer t, INSTR_Alloca t num_elems align)
         ] (* TODO: Generate atomic operations and other instructions *)
          ++ (if seq.nilp (filter_ptr_typs ctx) then [] else [gen_gep; gen_load; gen_store])
-         ++ (if seq.nilp (filter_agg_typs ctx) then [] else [gen_extractvalue; gen_insertvalue])).
-         (* ++ (if seq.nilp (filter_vec_typs ctx) then [] else [gen_extractelement; gen_insertelement])). *)
-
+         ++ (if seq.nilp (filter_agg_typs ctx) then [] else [gen_extractvalue; gen_insertvalue])
+         ++ (if seq.nilp (filter_vec_typs ctx) then [] else [gen_extractelement; gen_insertelement])).
+
```
Chobbes: Let's delete the trailing whitespace here (there's extra spaces on this line, it's annoying because it shows up in diffs)

Chobbes (changes requested):

src/coq/QC/GenAST.v
```diff
@@ -286,7 +286,12 @@ Section GenerationState.
        let new_ctx := vars ++ ctx in
        modify (replace_ctx new_ctx);;
        ret tt.
-
+  Definition hide_ctx {A} (g: GenLLVM A) : GenLLVM A
+    := ctx <- get_ctx;;
```
Chobbes: Looks good, I might recommend naming this `saved_ctx` or `old_ctx` or something.

src/coq/QC/GenAST.v
hanxic: I think I have a shorter way of writing it, but the reason I have this is that this function is used by `insertvalue` and `extractvalue`, and for convenience in the helper function  I included the prepath, which, because of what we put in as argument, will have one possible choice as nilpath. Therefore, we need to get rid of the nil path because `insertvalue` and `extractvalue` need at least one index.

src/coq/QC/GenAST.v
```diff
@@ -823,15 +829,33 @@ Fixpoint get_index_paths_aux (t_from : typ) (pre_path : list Z) {struct t_from}:
 Definition get_index_paths_ptr (t_from: typ) : list (typ * list (Z)) :=
   map (fun '(t, path) => (t, path)) (get_index_paths_aux t_from [0%Z]).
 
+
+(* Index path without getting into vector *)
+Fixpoint get_index_paths_aux_xinvec (t_from : typ) (pre_path : list Z) {struct t_from}: list (typ * list (Z)) :=
```
Chobbes: What do these names mean?

Chobbes (changes requested):

src/coq/QC/GenAST.v
hanxic: I was running out of names... Basically means get_index_paths_aux but don't get into vec... Please feel free to suggest other names

Chobbes:
Looks great, thank you!

PR #227, opened 2022-06-03
Added in insertValue insertElement and extractElement

nsobotka:
We added in functionality for these three functions in all three concretize_uvalue spots. Some of the test cases that I wrote for insert and extract value pass, but not all. None of the tests pass for insert and extract element, and I'm not yet sure if that's based on the test cases themselves or the code in Serialization.v

### DIFF
--- a/src/coq/Handlers/Serialization.v
+++ b/src/coq/Handlers/Serialization.v
@@ -141,12 +141,52 @@ Module Type SerializationBase (LP : LLVMParams) (MP : MemoryParams LP).
               | None => extractbytes_to_dvalue M undef_handler ERR_M lift_ue bytes dt
               end
             else extractbytes_to_dvalue M undef_handler ERR_M lift_ue bytes dt
+        | UVALUE_InsertValue t uv elt idxs =>
+            str <- concretize_uvalueM M undef_handler ERR_M lift_ue uv;;
+            elt <- concretize_uvalueM M undef_handler ERR_M lift_ue elt;;
+            let fix loop str idxs : ERR_M dvalue :=
+              match idxs with
+              | [] => ret str (* What to do? Do we print out some error *)
+              | i :: nil =>
+                  v <- insert_into_str str elt i;;
+                  ret v
+              | i :: tl =>
+                  v <- index_into_str_dv str i;;
+                  loop v tl
+              end in
+            lift_ue dvalue (loop str idxs)
+        | UVALUE_ExtractElement vec_typ vec idx =>
+            dvec <- concretize_uvalueM M undef_handler ERR_M lift_ue vec;;
+            didx <- concretize_uvalueM M undef_handler ERR_M lift_ue idx;;
+            let elt_typ := match dvec with
+                           | DVALUE_Vector elts => match elts with
+                                                  | [] => DTYPE_Void
+                                                  | hd::tl => match hd with
+                                                            | DVALUE_I1 _ => DTYPE_I 1
+                                                            | DVALUE_I8 _ => DTYPE_I 8
+                                                            | DVALUE_I32 _ => DTYPE_I 32
+                                                            | DVALUE_I64 _ => DTYPE_I 64
+                                                            | DVALUE_IPTR _ => DTYPE_Pointer
+                                                            | DVALUE_Float _ => DTYPE_Float
+                                                            | DVALUE_Double _ => DTYPE_Double
+                                                            | _ => DTYPE_Void (* Won't reach here *)
+                                                            end
+                                                  end
+                           | _ => DTYPE_Void (* Won't reach here*)
+                           end in
+            lift_ue dvalue (index_into_vec_dv elt_typ dvec didx)
+        | UVALUE_InsertElement vec_typ vec elt idx =>
+            dvec <- concretize_uvalueM M undef_handler ERR_M lift_ue vec;;
+            didx <- concretize_uvalueM M undef_handler ERR_M lift_ue idx;;
+            delt <- concretize_uvalueM M undef_handler ERR_M lift_ue elt;;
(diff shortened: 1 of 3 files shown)


src/coq/Handlers/Serialization.v
```diff
@@ -874,7 +914,45 @@ Module MakeBase (LP : LLVMParams) (MP : MemoryParams LP) : SerializationBase LP
...
+            str <- concretize_uvalueM uv;;
+            elt <- concretize_uvalueM elt;;
+            let fix loop str idxs : ERR_M dvalue :=
+              match idxs with
+              | [] => ret str (* What to do? Do we print out some error *)
+              | i :: nil =>
+                  v <- insert_into_str str elt i;;
+                  ret v
+              | i :: tl =>
+                  v <- index_into_str_dv str i;;
```
Chobbes: This doesn't seem right. It looks like you'll do an insertion for each index in `idxs`, but `idxs` is just a path for where you insert a single value into an aggregate type, right?

src/coq/Handlers/Serialization.v
```diff
@@ -874,7 +914,45 @@ Module MakeBase (LP : LLVMParams) (MP : MemoryParams LP) : SerializationBase LP
...
+                  loop v tl
+              end in
+            lift_ue (loop str idxs)
+        | UVALUE_ExtractElement vec_typ vec idx =>
+            dvec <- concretize_uvalueM vec;;
+            didx <- concretize_uvalueM idx;;
+            let elt_typ := match dvec with
```
Chobbes: You should use `vec_typ` to get this. `vec_typ` is the type of `vec`.

Chobbes (changes requested):

src/coq/Handlers/Serialization.v
nsobotka: It only inserts when we are at the last index, otherwise it indexes down a layer

Chobbes: Oh, I see. You're right about that, I misread `index` as `insert` (ugh).

However, I still think this is a mistake, then, no?

If you `index_into_str_dv` you're getting a field from a structure or whatever, so you will end up returning something with the wrong type. You're throwing away the other fields of the structure every time you index, no?

src/coq/Handlers/Serialization.v
```diff
@@ -141,12 +141,42 @@ Module Type SerializationBase (LP : LLVMParams) (MP : MemoryParams LP).
...
+              match idxs with
+              | [] => ret str (* What to do? Do we print out some error *)
+              | i :: nil =>
+                  v <- insert_into_str str elt i;;
+                  ret v
+              | i :: tl =>
+                      v <- index_into_str_dv str i;;
+                      v2 <- loop v tl;;
+                      v3 <- insert_into_str str v2 i;;
+                      ret v3
```
Chobbes: Let's call:

- `v` should be `subfield`
- `v2` can be `modified_subfield`

And then let's just kill `v3` and instead of

```
v3 <- insert_into_str str modified_subfield i;;
ret v3
```

We can just have

```
insert_into_str str modified_subfield i
```

src/coq/Handlers/Serialization.v
```diff
@@ -141,12 +141,42 @@ Module Type SerializationBase (LP : LLVMParams) (MP : MemoryParams LP).
...
+              end in
+            lift_ue dvalue (loop str idxs)
+        | UVALUE_ExtractElement vec_typ vec idx =>
+            dvec <- concretize_uvalueM M undef_handler ERR_M lift_ue vec;;
+            didx <- concretize_uvalueM M undef_handler ERR_M lift_ue idx;;
+            let elt_typ := match vec_typ with
```
Chobbes: Made another comment, but let's raise an error here instead of introducing a garbage value that might break things in unexpected ways.

src/coq/Handlers/Serialization.v
```diff
@@ -141,12 +141,42 @@ Module Type SerializationBase (LP : LLVMParams) (MP : MemoryParams LP).
...
+            didx <- concretize_uvalueM M undef_handler ERR_M lift_ue idx;;
+            delt <- concretize_uvalueM M undef_handler ERR_M lift_ue elt;;
+            lift_ue dvalue (insert_into_vec_dv vec_typ dvec delt didx)
         | _ =>
             lift_ue dvalue
                     (raise_error
                        ("concretize_uvalueM: Attempting to convert a partially non-reduced uvalue to dvalue. Should not happen: " ++
                                                                                                                                   uvalue_constructor_string u))
         end.
+  Check concretize_uvalueM.
```
Chobbes: Delete this.

src/coq/Handlers/Serialization.v
```diff
@@ -141,12 +141,42 @@ Module Type SerializationBase (LP : LLVMParams) (MP : MemoryParams LP).
               | None => extractbytes_to_dvalue M undef_handler ERR_M lift_ue bytes dt
               end
             else extractbytes_to_dvalue M undef_handler ERR_M lift_ue bytes dt
+        | UVALUE_InsertValue t uv elt idxs =>
+            str <- concretize_uvalueM M undef_handler ERR_M lift_ue uv;;
+            elt <- concretize_uvalueM M undef_handler ERR_M lift_ue elt;;
+            let fix loop str idxs : ERR_M dvalue :=
+              match idxs with
+              | [] => ret str (* What to do? Do we print out some error *)
```
Chobbes: This is a good point, let's make this raise an error instead of failing silently and terrifyingly.

Chobbes (changes requested):

Chobbes:
Looks good to me, thank you!

PR #238, opened 2022-06-07
Update Makefile to not build QC.

nsobotka:
Makefile updates (done by Prof. Zdancewic) and some more test cases.


Chobbes:
In the future let's give the PRs a descriptive name.

Issue #223, opened 2022-06-01
Fix Alloca with double pointers


Chobbes:
I believe this is only a performance issue... When generating an alloca we currently use `gen_sized_typ`...

```
t <- gen_sized_typ;;
num_elems <- ret None;;
align <- ret None;;
ret (TYPE_Pointer t, INSTR_Alloca t num_elems align)
```

It is possible for `t` to be a pointer itself. We require all `alloca`s to be followed with a store to initialize the freshly allocated memory. The problem is that we may not be able to find a pointer with the appropriate type in the context to initialize with.

The initialization is done with `fix_alloca` which calls:

```
  Definition gen_store_to (ptr : texp typ) : GenLLVM (typ * instr typ)
    :=
      match ptr with
      | (TYPE_Pointer t, pexp) =>
        vol   <- lift arbitrary;;
        align <- ret None;;

        e <- resize_LLVM 0 (gen_exp t);;
        let val := (t, e) in

        ret (TYPE_Void, INSTR_Store vol val ptr align)
      | _ => lift failGen
      end.
```

`e` will be generated by looping for an identifier in the context that matches the pointer type. If it's not there it will call `failGen`, causing us not to generate a test case (which is a performance issue).

One potential solution:

- Replace `gen_sized_typ` when generating an `alloca` with a similar function that will only generate pointer types that already exist in the context.

I don't think this is a huge performance impact, but might be relevant to #221

PR #241, opened 2022-06-08
Memory testing gen: ptrtoint and inttoptr generator

### LINKED ISSUE #212: Generate inttoptr and ptrtoint casts
Generate programs that perform `inttoptr` and `ptrtoint` casts. This will likely involve extending the `GenState` structure used by `GenLLVM` in order to:

- keep track of bounds so we know when an `int` can be cast back to a `ptr`.
- keep track of the type the pointer had originally. We need to at least make sure that when we cast back to a pointer we don't get a pointer to a type with a larger size that would result in an out of bounds read.

hanxic:
PR for ptrtoint and inttoptr. Can make and generate both generators. The only case not covered yet is when we want to generate a vector inside the pointer returned from inttoptr (went stack overflow). Shouldn't be a difficult fix but need to give some thought.

### DIFF
--- a/src/coq/QC/GenAST.v
+++ b/src/coq/QC/GenAST.v
@@ -151,6 +151,7 @@ Section GenerationState.
     ; gen_ctx : list (ident * typ)
     (* Type aliases *)
     ; gen_typ_ctx : list (ident * typ)
+    ; gen_ptrtoint_ctx : list (typ * ident * typ)
     }.
 
   Definition init_GenState : GenState
@@ -160,6 +161,7 @@ Section GenerationState.
         ; num_blocks := 0
         ; gen_ctx    := []
         ; gen_typ_ctx    := []
+        ; gen_ptrtoint_ctx := []
        |}.
 
   Definition increment_raw (gs : GenState) : GenState
@@ -169,6 +171,7 @@ Section GenerationState.
         ; num_blocks  := gs.(num_blocks)
         ; gen_ctx     := gs.(gen_ctx)
         ; gen_typ_ctx := gs.(gen_typ_ctx)
+        ; gen_ptrtoint_ctx := gs.(gen_ptrtoint_ctx)
        |}.
 
   Definition increment_global (gs : GenState) : GenState
@@ -178,6 +181,7 @@ Section GenerationState.
         ; num_blocks  := gs.(num_blocks)
         ; gen_ctx     := gs.(gen_ctx)
         ; gen_typ_ctx := gs.(gen_typ_ctx)
+        ; gen_ptrtoint_ctx := gs.(gen_ptrtoint_ctx)
        |}.
 
   Definition increment_void (gs : GenState) : GenState
@@ -187,6 +191,7 @@ Section GenerationState.
         ; num_blocks  := gs.(num_blocks)
         ; gen_ctx     := gs.(gen_ctx)
         ; gen_typ_ctx := gs.(gen_typ_ctx)
+        ; gen_ptrtoint_ctx := gs.(gen_ptrtoint_ctx)
        |}.
 
   Definition increment_blocks (gs : GenState) : GenState
@@ -196,6 +201,7 @@ Section GenerationState.
         ; num_blocks  := S gs.(num_blocks)
         ; gen_ctx     := gs.(gen_ctx)
         ; gen_typ_ctx := gs.(gen_typ_ctx)
+        ; gen_ptrtoint_ctx := gs.(gen_ptrtoint_ctx)
        |}.
 
   Definition replace_ctx (ctx : list (ident * typ)) (gs : GenState) : GenState
@@ -205,6 +211,7 @@ Section GenerationState.
         ; num_blocks  := gs.(num_blocks)
         ; gen_ctx     := ctx
         ; gen_typ_ctx := gs.(gen_typ_ctx)
+        ; gen_ptrtoint_ctx := gs.(gen_ptrtoint_ctx)
        |}.
 
   Definition replace_typ_ctx (typ_ctx : list (ident * typ)) (gs : GenState) : GenState
@@ -214,8 +221,18 @@ Section GenerationState.
         ; num_blocks  := gs.(num_blocks)
         ; gen_ctx     := gs.(gen_ctx)
         ; gen_typ_ctx := typ_ctx
+        ; gen_ptrtoint_ctx := gs.(gen_ptrtoint_ctx)
        |}.
 
+  Definition replace_ptrtoint_ctx (ptrtoint_ctx : list (typ * ident * typ)) (gs: GenState) : GenState
+    := {| num_void    := gs.(num_void)
+        ; num_raw     := gs.(num_raw)
+        ; num_global  := gs.(num_global)
+        ; num_blocks  := gs.(num_blocks)
+        ; gen_ctx     := gs.(gen_ctx)
+        ; gen_typ_ctx := gs.(gen_typ_ctx)
+        ; gen_ptrtoint_ctx := ptrtoint_ctx
+       |}.
   Definition GenLLVM := stateT GenState G.
 
   (* Need this because extlib doesn't declare this instance as global :|. *)
@@ -269,6 +286,9 @@ Section GenerationState.
   Definition get_typ_ctx : GenLLVM (list (ident * typ))
     := gets (fun gs => gs.(gen_typ_ctx)).
 
+  Definition get_ptrtoint_ctx : GenLLVM (list (ident * typ))
+    := gets (fun gs => gs.(gen_ptrtoint_ctx)).
+  
   Definition add_to_ctx (x : (ident * typ)) : GenLLVM  unit
     := ctx <- get_ctx;;
        let new_ctx := x :: ctx in
@@ -279,6 +299,12 @@ Section GenerationState.
     := ctx <- get_typ_ctx;;
        let new_ctx := x :: ctx in
        modify (replace_typ_ctx new_ctx);;
+       ret tt.
+
+  Definition add_to_ptrtoint_ctx (x : (ident * typ)) : GenLLVM unit
+    := ctx <- get_ptrtoint_ctx;;
+       let new_ctx := x :: ctx in
+       modify (replace_ptrtoint_ctx ctx);;
        ret tt.
 
   Definition append_to_ctx (vars : list (ident * typ)) : GenLLVM unit
@@ -939,7 +965,6 @@ Definition gen_extractelement : GenLLVM (typ * instr typ) :=
   ret (t_in_vec, INSTR_Op (OP_ExtractElement (tvec, EXP_Ident id) (TYPE_I 32%N, EXP_Integer (Z.of_nat (N.to_nat index_for_extractelement))))).
 
 (* Assumes input is a primitive type*)
-
 (* Generate the element to be inserted into vector*)
 Definition gen_typ_eq_prim_typ (t: typ): GenLLVM (exp typ) :=
   match t with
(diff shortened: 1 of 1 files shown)


hanxic:
Have stuff to fix... There are some bugs I found

hanxic:
Bug fixed. Still cannot generate vector. Everytime I move the prime type out to its own function it will give me stack overflow really quick

src/coq/QC/GenAST.v
```diff
@@ -964,6 +999,88 @@ Definition gen_insertelement : GenLLVM (typ * instr typ) :=
...
+  ctx <- get_ptrtoint_ctx;;
+  '(old_tptr, id, typ_from_cast) <- oneOf_LLVM (map ret ctx);;
+  let genllvm_new_tptr :=
+    match old_tptr with
+    | TYPE_Pointer old_typ => new_typ <- gen_typ_le_size (get_size_from_typ old_typ);;
+                                           ret (TYPE_Pointer new_typ)
+    | TYPE_Vector sz old_tptr => match old_tptr with
+                                | TYPE_Pointer old_typ =>
+                                    new_typ <-gen_typ_le_size (get_size_from_typ old_typ);;
+                                    ret (TYPE_Pointer new_typ)
```
Chobbes: This seems a bit wrong to me.

When you use `ptrtoint` on a vector of pointers, you will get a vector of integers.

When you use `inttoptr` on a vector of integers, you should get a vector of pointers back.

The fact that we return a `TYPE_Pointer` here seems dubious to me. You can't combine everything that's in the vector into one pointer.

src/coq/QC/GenAST.v
```diff
@@ -964,6 +999,88 @@ Definition gen_insertelement : GenLLVM (typ * instr typ) :=
...
+(*TODO: code for ptrtoint incomplete*)
+Definition filter_ptr_vecptr_typ (ctx: list (ident * typ)) : list (ident * typ) :=
+  filter (fun '(_, t) => match t with
+                | TYPE_Pointer _ => true
+                | TYPE_Vector _ ty => match ty with
+                                     | TYPE_Pointer _ => true
+                                     | _ => false
+                                     end
+                | _ => false
+                      end) ctx.
```
Chobbes: Add a newline between functions so it's easier to pick them out. Otherwise the function bodies and definitions kind of blend together.

Chobbes (changes requested):

src/coq/QC/GenAST.v
hanxic: Yes that's a bug. Should be vector of pointer

src/coq/QC/GenAST.v
```diff
@@ -214,8 +221,18 @@ Section GenerationState.
...
+  Definition replace_ptrtoint_ctx (ptrtoint_ctx : list (typ * ident * typ)) (gs: GenState) : GenState
+    := {| num_void    := gs.(num_void)
+        ; num_raw     := gs.(num_raw)
+        ; num_global  := gs.(num_global)
+        ; num_blocks  := gs.(num_blocks)
+        ; gen_ctx     := gs.(gen_ctx)
+        ; gen_typ_ctx := gs.(gen_typ_ctx)
+        ; gen_ptrtoint_ctx := ptrtoint_ctx
+       |}.
   Definition GenLLVM := stateT GenState G.
```
Chobbes, line 237: Newline :).

Chobbes (changes requested):

src/coq/QC/GenAST.v
```diff
@@ -964,6 +999,88 @@ Definition gen_insertelement : GenLLVM (typ * instr typ) :=
...
+  | TYPE_Packed_struct fields => fold_right (fun x acc => ((get_size_from_typ x) + acc)%nat) 0%nat fields
+  | TYPE_Pointer _ => 64
+  | _ => 1
+  end.
+
+Fixpoint gen_typ_le_size (max_byte_sz : nat) : GenLLVM typ :=
+  ctx <- get_ctx;;
+  oneOf_LLVM ((if (max_byte_sz =? 0)%nat then [] else
+               [ret (TYPE_I 1)]
+                ++ (if (8 <=? max_byte_sz)%nat then [ret (TYPE_I 8); x <- gen_typ_le_size max_byte_sz;;ret (TYPE_Pointer x)] else [])
```
Chobbes: For testing we should assume that pointers are 8 bytes wide, I think. They're the same size as an i64. Also a pointer to *any* type is the same size, because a pointer is just a location in memory. That means you can technically make `x` any type, instead of one of a restricted size....

HOWEVER. I don't think we should generate pointers for this, because that means we will cast random integer values etc to pointers, and that location in memory won't be allocated (so when we go to generate loads and stores to those pointers things will explode).

Technically we should make this a parameter because it depends on the memory model, but we can fix the size for now and maybe leave a note + make a github issue about it.

Chobbes (changes requested):

src/coq/QC/GenAST.v
hanxic: I just want to clarify. So are you saying that I should comment out the pointer in this function, for now

Chobbes: Yeah, I think so. Generating pointers here will cause problems.

hanxic: Commented out already in the newest push

Chobbes:
I think there's a problem still. We will generate a `inttoptr` cast whenever there is something in the `ptrtoint_ctx`. That's fine, *BUT*...

If the type in the context contains a second pointer anywhere (double pointers, structures with pointers, etc), then we should be much more careful. We can't cast to any type with size less than or equal to the original type --- those pointers *MUST* still be pointers. The easy way to handle this for now is to detect when there are pointers contained within the type, and only cast back to the original type instead.

hanxic:
Cleaned up and solve the problem (I think). I added additional match in `gen_inttoptr` that will check if it is double pointer. If not we will generate new type.

src/coq/QC/GenAST.v
hanxic: Sorry. I really need to be more careful on this. Added

src/coq/QC/GenAST.v
```diff
@@ -964,6 +1000,118 @@ Definition gen_insertelement : GenLLVM (typ * instr typ) :=
...
+       ret ([subtyp] ++ tl).
+
+Definition gen_inttoptr : GenLLVM (typ * instr typ) :=
+  ctx <- get_ptrtoint_ctx;;
+  '(old_tptr, id, typ_from_cast) <- oneOf_LLVM (map ret ctx);;
+  let genllvm_new_tptr :=
+    match old_tptr with
+    | TYPE_Pointer old_typ =>
+        match old_typ with
+        | TYPE_Pointer _ => (* Double pointer issue *)
```
Chobbes: I mentioned this in slack, but just to keep it all in one place...

We're going to have to make sure that there are no nested pointers in `old_typ`. That includes double pointers, triple pointers, pointers to structs that contain pointers, structs that contain structs that contain structs that contain arrays that contain pointers, etc... You'll need a recursive function (maybe called something like `typ_contains_pointer`) to check if there's a pointer within a type, can't just be a simple match like this.

Chobbes (changes requested):

src/coq/QC/GenAST.v
hanxic: Oh ic. So sorry. I simplified the problem. Working on it now. And just to clarify, if there are things like {i32, i32, i32*}, am I able to change i32 into something else and keep i32* or just leave the entire structure unchanged

Hchen201205: nvm I see your answer above. Pointer inside the struct will remains, with rest of them changing. Got it

hanxic: Got it. I will write a recursive function that will see if a type has a pointer, and then will cast if not

hanxic:
Implemented typ_contains_pointers that will see if the old type (pointer, or vector of pointers, the only two possible input for ptrtoint) contains any double-pointer. Add some catches in generator that allows 0 size array, and not allow 0 size vector.
Not removing `ctx <- get_ctx` will prevent the stack overflow issue. Not sure why delete that will not.

hanxic:
The force-push happened because there is a solution for #223 that I was supposed to do in another branch...

src/coq/QC/GenAST.v
```diff
@@ -964,6 +1003,139 @@ Definition gen_insertelement : GenLLVM (typ * instr typ) :=
...
+            end) ctx.
+
+Definition gen_ptrtoint : GenLLVM (typ * instr typ) :=
+  ctx <- get_ctx;;
+  let ptr_in_context := filter_ptr_typs ctx in
+  '(id, tptr) <- (oneOf_LLVM (map ret ptr_in_context));;
+  let gen_typ_in_ptr :=
+    match tptr with
+    | TYPE_Pointer t =>
+        gen_int_typ
```
Chobbes: HMMMM. I'm realizing that this will actually be a problem.

If you cast a pointer to, say, an `i1`, `i8`, `i32`, or even an `i64` that integer data type may not have enough space to hold the pointer value (technically we have unlimited pointers). In practice `i32` and `i64` conversions should work, but I think we should actually only cast to `TYPE_IPTR` for now. I'd make a `gen_int_typ_for_ptr_cast` (so we can update it later) that you'll use instead of `gen_int_typ` here. It should only return `TYPE_IPTR` for now.

Otherwise if you cast to an `i8` you might truncate the pointer value (or if you cast to an `i1` you almost certainly will). For instance if I allocate something and it gets address `300` an `i8` can't fit that and will truncate the value to `44`... When you go to cast `44` back to a pointer you won't end up pointing to address `300`, you'll be pointing somewhere else. Uh-oh!

This might not be caught with `vellvm` raising an error because we only do `alloca`s to allocate memory, and since we only have one function `main` we never return from a function to pop the stack and free any `alloca`s. Since memory is allocated in order, all lower addresses will be allocated... So when you access `44` it will actually find something and overwrite it. Normally this would be caught with a provenance mismatch, but integers cast to pointers are allowed to accesss any memory.

THAT SAID it should give different behaviour with clang if you end up writing / loading to different points in memory because of this (which you would). I'd be curious why you're not getting a test case failure --- might be good to look into. I'd expect one whenever you cast to an `i1`, so try replacing `gen_int_typ` here with `ret (TYPE_I 1)`... I'd still expect clang to differ on an `i8`, but you won't get any truncation unless you allocate more than 256 bytes in the LLVM program, which we might not because the programs we generate are small.

Chobbes (changes requested):

src/coq/QC/GenAST.v
hanxic: I can change that, but shouldn't I also change the show instance. The reason is currently it is `TYPE_IPTR => "iptr"`, which I believe will not pass the syntax check. I can change it to just return "i32*"

hanxic: Also I thought you are allowed to truncate pointer? In Langref it said "if value is larger than ty2, the type we will convert to, then a truncation is done."

src/coq/QC/GenAST.v
```diff
@@ -551,7 +588,11 @@ Section TypGenerators.
                 ; TYPE_I 32
                 (* TODO: big ints *)
                 (* ; TYPE_I 64 *)
-                ]).
+           ]).
+
+  (* TODO: IPTR not implemented *)
+  (* Definition gen_int_typ_for_ptr_cast : GenLLVM typ :=
```
Chobbes: I think we should still use this function below, just make this return an `i64` instead of an `iptr`. Then if we want to add support for casting to different integer types we have one place we have to change here, and it's a nice named function. Other than that these changes are looking good, and I can get them merged.

Chobbes (changes requested):

src/coq/QC/GenAST.v
Chobbes: You can truncate a pointer, it's just that it would be bad if we truncate it and then try to cast it back to a pointer, because we lose information and can't recreate the original pointer. It's valid to truncate, but we have to be careful to make sure we only produce valid programs, and with how we generate things currently if we truncate we may access unallocated memory.

Issue #221, opened 2022-06-01
QuickChick Performance


Chobbes:
There seems to be a couple of performance issues with the QC generator. Here's some things I've noticed:

- [x] We can't seem to generate large code blocks --- I think `failGen` is getting called somewhere: `Sample (run_GenLLVM (gen_code_length 10)).`
- [ ]  After some fixes to the generator it now seems to get stuck when generating larger programs. This could be because of a loop in one of the generators? I don't think it can just be calling `failGen` constantly, because I think it will just stop testing after a while, but that might be another possibility.
- [x] Some things don't seem to be generated... Like the `insert` / `extract` instructions. Could be related to the performance issues here.

Chobbes:
Some of the issues have been addressed with this: 3fc4ad23c12d121734e4ad18702a0fe07d0411c4

There was a problem with gen_load / gen_store which made them much more likely to fail, especially once we started generating aggregate types.

Chobbes:
#223 is relevant, but I believe it should only be a minor concern.

Chobbes:
Another thing we should tune in the future: 

- [ ] The probability of generating a type that's already in the context. This is partially being addressed in #246, but we haven't spent much time tuning how likely it should be to reuse something in the context.

Chobbes:
@Hchen201205 discovered that `gen_loop_sz` seems to be the cause of our infinite looping woes. It turns out we weren't writing the `temporary_vellvm.ll` file before running the `vellvm` interpreter on the program, so we wouldn't see a program that would actually get stuck. Swapping the order that we test things in we found this test case:

```
define i8 @main() {
b0:
    %v0 = alloca [1 x <{[3 x <6 x float>], [5 x {float, <{i8, i1, i32, i1, i32}>, {float, i8, float}, <{i1, float, i1, i32}>, i8}]}>]
[...]

PR #246, opened 2022-06-09
Memory testing performance: PR #223 on alloca performance

### LINKED ISSUE #223: Fix Alloca with double pointers

hanxic:
Switched `gen_sized_typ` with a new function `gen_sized_typ_ptrinctx` that does a similar thing except has the possibility of generating a pointer only when there is a pointer in the ctx.

### DIFF
--- a/src/coq/QC/GenAST.v
+++ b/src/coq/QC/GenAST.v
@@ -397,18 +397,18 @@ Definition filter_sized_typs (typ_ctx: list (ident * typ)) (ctx : list (ident * 
     | (S sz') =>
         ctx <- get_ctx;;
         aliases <- get_typ_ctx;;
-        let typs_in_ctx := map (fun '(_, typ) => ret typ) (filter_sized_typs aliases ctx) in
-        oneOf_LLVM
-        (typs_in_ctx ++[gen_sized_typ_0
-        ; ret TYPE_Pointer <*> gen_sized_typ_size sz'
+        let typs_in_ctx := map (fun '(_, typ) => (1%nat, ret typ)) (filter_sized_typs aliases ctx) in
+       freq_LLVM
+        (typs_in_ctx ++[(1%nat, gen_sized_typ_0)
+        ; (1%nat, ret TYPE_Pointer <*> gen_sized_typ_size sz')
         (* TODO: Might want to restrict the size to something reasonable *)
-        ; ret TYPE_Array <*> lift_GenLLVM genN <*> gen_sized_typ_size sz'
-        ; ret TYPE_Vector <*> (n <- lift_GenLLVM genN;; ret (n + 1)%N) <*> gen_sized_typ_size 0
+        ; (1%nat, ret TYPE_Array <*> lift_GenLLVM genN <*> gen_sized_typ_size sz')
+        ; (1%nat, ret TYPE_Vector <*> (n <- lift_GenLLVM genN;; ret (n + 1)%N) <*> gen_sized_typ_size 0)
         (* TODO: I don't think functions are sized types? *)
         (* ; let n := Nat.div sz 2 in *)
         (*   ret TYPE_Function <*> gen_sized_typ_size n <*> listOf_LLVM (gen_sized_typ_size n) *)
-        ; ret TYPE_Struct <*> nonemptyListOf_LLVM (gen_sized_typ_size sz')
-        ; ret TYPE_Packed_struct <*> nonemptyListOf_LLVM (gen_sized_typ_size sz')
+        ; (1%nat, ret TYPE_Struct <*> nonemptyListOf_LLVM (gen_sized_typ_size sz'))
+        ; (1%nat, ret TYPE_Packed_struct <*> nonemptyListOf_LLVM (gen_sized_typ_size sz'))
           ])
     end.
   Next Obligation.
@@ -424,17 +424,17 @@ Definition filter_sized_typs (typ_ctx: list (ident * typ)) (ctx : list (ident * 
     | (S sz') =>
         ctx <- get_ctx;;
         aliases <- get_typ_ctx;;
-        let typs_in_ctx := map (fun '(_, typ) => ret typ) (filter_sized_typs aliases ctx) in
-      oneOf_LLVM
-        (typs_in_ctx ++ [ gen_sized_typ_0
+        let typs_in_ctx := map (fun '(_, typ) => (1%nat, ret typ)) (filter_sized_typs aliases ctx) in
+      freq_LLVM
+        (typs_in_ctx ++ [(1%nat, gen_sized_typ_0)
         (* TODO: Might want to restrict the size to something reasonable *)
-        ; ret TYPE_Array <*> lift_GenLLVM genN <*> gen_sized_typ_size_ptrinctx sz'
-        ; ret TYPE_Vector <*> (n <- lift_GenLLVM genN;; ret (n + 1)%N) <*> gen_sized_typ_size_ptrinctx 0
+        ; (1%nat, ret TYPE_Array <*> lift_GenLLVM genN <*> gen_sized_typ_size_ptrinctx sz')
+        ; (1%nat, ret TYPE_Vector <*> (n <- lift_GenLLVM genN;; ret (n + 1)%N) <*> gen_sized_typ_size_ptrinctx 0)
         (* TODO: I don't think functions are sized types? *)
         (* ; let n := Nat.div sz 2 in *)
         (*   ret TYPE_Function <*> gen_sized_typ_size n <*> listOf_LLVM (gen_sized_typ_size n) *)
-        ; ret TYPE_Struct <*> nonemptyListOf_LLVM (gen_sized_typ_size_ptrinctx sz')
-        ; ret TYPE_Packed_struct <*> nonemptyListOf_LLVM (gen_sized_typ_size_ptrinctx sz')])
+        ; (1%nat, ret TYPE_Struct <*> nonemptyListOf_LLVM (gen_sized_typ_size_ptrinctx sz'))
+        ; (1%nat, ret TYPE_Packed_struct <*> nonemptyListOf_LLVM (gen_sized_typ_size_ptrinctx sz'))])
     end.
   Next Obligation.
   lia.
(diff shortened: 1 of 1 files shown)


src/coq/QC/GenAST.v
```diff
@@ -404,6 +412,33 @@ Section TypGenerators.
...
+        ; ret TYPE_Vector <*> (n <- lift_GenLLVM genN;; ret (n + 1)%N) <*> gen_sized_typ_size_ptrinctx 0
+        (* TODO: I don't think functions are sized types? *)
+        (* ; let n := Nat.div sz 2 in *)
+        (*   ret TYPE_Function <*> gen_sized_typ_size n <*> listOf_LLVM (gen_sized_typ_size n) *)
+        ; ret TYPE_Struct <*> nonemptyListOf_LLVM (gen_sized_typ_size_ptrinctx sz')
+        ; ret TYPE_Packed_struct <*> nonemptyListOf_LLVM (gen_sized_typ_size_ptrinctx sz')
+
+        ; ctx <- get_ctx;;
+          let ptrs_in_ctx := filter_ptr_typs ctx in
+          if seq.nilp ptrs_in_ctx then '(_,typ) <- oneOf_LLVM (map ret ctx);; ret typ else gen_sized_typ_size_ptrinctx sz' (* If not simply wait till the next size. This will prevent any failGen *)
```
Chobbes: Hmmmmmmm.

Why call `filter_ptr_typs` if you're just going to pick from anything in the regular context? Picking from the regular context actually isn't a bad idea --- it makes it more likely that we pick a type of an already existing variable. We should probably use `freq_LLVM` to always do that with some weighted probability here and in `gen_sized_typ`. You get more interesting test cases when you reuse variables as parts of the program intermingle more, if that makes sense.

The other thing is I think this would be more efficient if we did the thing where we conditionally append this generator into the list if the context isn't empty, instead of calling it recursively.

Chobbes (changes requested):

src/coq/QC/GenAST.v
hanxic: Got it. I didn't know whether I am supposed to use everything in ctx, which is why I use ctx only for one case. I will start changing it to everything in ctx, I will give it a low frequency compared to others so that we won't largely rely on ctx.

I had some trouble, algorithmically, on the second thing you said because everytime I process something in GenLLVM realm I need to return GenLLVM. I do think I may have a solution now. Will update tonight

Chobbes: I just mean normally we do the trick where we say...

```
oneOf_LLVM ([gen1; gen2] ++ (if there_is_something_in_ctx then [ gen3 ] else [ ]))
```
The advantage is we won't even select `gen3` when it's not meaningful.

Chobbes: I think it should actually have a fairly high frequency to pick a type that already exists in the context (assuming the context is non-empty). This would make it more likely that we generate programs that actually use variables defined earlier --- instead of creating a bunch of variables that are never used (and never have an impact upon the rest of the execution of the program).

We could even make this frequency clever --- increasing the frequency that we'll reuse a type from the context when the context is larger (i.e., when we've already got a bunch of types in our program, we can make it less likely to generate new ones). This could be a good idea to stop our generators from just generating distinct types of things all the time that can't interopt. I'll make a new issue for this, though.

The other thing is... You just pull `typ` out of `ctx`, but `ctx` may have non-sized types in it, so this is a bug!

hanxic: I was using the trick you mentioned but the problem with that is, whenever you do something related to the ctx, you have to return a GenLLVM in the helper. Therefore it has some complication there, I think I have some ideas to do it though.

The context large enough thing shouldn't be a big problem. I can have a dummy number there and we can see what large enough is enough.

Can you elaborate on what is the non-sized types? Are you saying that it's illegal because it's size (number of layers in this context) is not the same as what we want?

Chobbes: Let's just clean this up to do the right thing first, and then worry about #247.

> I was using the trick you mentioned but the problem with that is, whenever you do something related to the ctx, you have to return a GenLLVM in the helper. Therefore it has some complication there, I think I have some ideas to do it though.

I don't follow, it's the same as everywhere else we do this, no?

> Can you elaborate on what is the non-sized types? Are you saying that it's illegal because it's size (number of layers in this context) is not the same as what we want?

Not all types are sized types (e.g., void, function types, etc). I think currently we don't put any such things in the context yet, but we could put functions in the context when we generate them.

hanxic:
Complete adding sized-type in context. I added it to all methods (`gen_typ_non_void_size_wo_fn`...) please let me know if I need to take it out from some of them. 

Also, I only start consider types in ctx if the size is greater than 0. Let me know if that needs to change.

Currently adding everything in ctx is a way of letting it to generate more frequent from ctx if ctx is larger. Will change it to freq_LLVM if we want more complicated probability. One way to do it is to map and let each type of ctx to all have same natural number (frequency) that is greater than 1, therefore make it more likely to generate from ctx when ctx is larger. 

Everything should compile.

src/coq/QC/GenAST.v
```diff
@@ -439,17 +478,21 @@ Section TypGenerators.
...
-                      ; ret TYPE_Vector <*> (n <- lift_GenLLVM genN;;ret (n + 1)%N) <*> gen_sized_typ_size 0
-                      ; let n := Nat.div sz 2 in
-                        ret TYPE_Function <*> gen_typ_size n <*> listOf_LLVM (gen_sized_typ_size n)
-                      ; ret TYPE_Struct <*> nonemptyListOf_LLVM (gen_sized_typ_size sz')
-                      ; ret TYPE_Packed_struct <*> nonemptyListOf_LLVM (gen_sized_typ_size sz')
-                      ]
+    | (S sz') =>
+        ctx <- get_ctx;;
+        aliases <- get_typ_ctx;;
+        let typs_in_ctx := map (fun '(_, typ) => ret typ) (filter_sized_typs aliases ctx) in
```
Chobbes: I don't think this should filter sized types.

Chobbes (changes requested):

src/coq/QC/GenAST.v
```diff
@@ -404,6 +418,31 @@ Section TypGenerators.
...
 
+ Program Fixpoint gen_sized_typ_size_ptrinctx (sz : nat) {measure sz} : GenLLVM typ :=
+    match sz with
+    | 0%nat => gen_sized_typ_0
+    | (S sz') =>
+        ctx <- get_ctx;;
+        aliases <- get_typ_ctx;;
+        let typs_in_ctx := map (fun '(_, typ) => ret typ) (filter_sized_typs aliases ctx) in
+      oneOf_LLVM
+        (typs_in_ctx ++ [ gen_sized_typ_0
```
Chobbes: We should use `freq_LLVM` instead of just relying on the size of `types_in_ctx`.

Chobbes (changes requested):

src/coq/QC/GenAST.v
hanxic: I'm assuming you are saying for this function specifically we don't need to filter the types right? I get that because this function uses void and function type.

src/coq/QC/GenAST.v
```diff
@@ -493,31 +536,39 @@ Section TypGenerators.
...
-                      ; ret TYPE_Vector <*> (n <- lift_GenLLVM genN;;ret (n + 1)%N) <*> gen_sized_typ_size 0
-                      ; let n := Nat.div sz 2 in
-                        ret TYPE_Function <*> gen_typ_size n <*> listOf_LLVM (gen_sized_typ_size n)
-                      ; ret TYPE_Struct <*> nonemptyListOf_LLVM (gen_sized_typ_size sz')
-                      ; ret TYPE_Packed_struct <*> nonemptyListOf_LLVM (gen_sized_typ_size sz')
-                      ]
+    | (S sz') =>
+        ctx <- get_ctx;;
+        aliases <- get_typ_ctx;;
+        let typs_in_ctx := map (fun '(_, typ) => (1%nat, ret typ)) (filter_sized_typs aliases ctx) in
```
Chobbes: We should probably use a `filter_non_void_typs` or something. I'm not 100% sure if they'll end up being the same long term, but I think it'd be good to make the distinction for now.

src/coq/QC/GenAST.v
```diff
@@ -404,6 +418,31 @@ Section TypGenerators.
...
     := sized_LLVM (fun sz => gen_sized_typ_size sz).
 
+ Program Fixpoint gen_sized_typ_size_ptrinctx (sz : nat) {measure sz} : GenLLVM typ :=
+    match sz with
+    | 0%nat => gen_sized_typ_0
+    | (S sz') =>
+        ctx <- get_ctx;;
+        aliases <- get_typ_ctx;;
+        let typs_in_ctx := map (fun '(_, typ) => (1%nat, ret typ)) (filter_sized_typs aliases ctx) in
+      freq_LLVM
```
Chobbes: So, I think we should use `freq_LLVM` like...

`freq_LLVM ([(min (length typs_in_ctx) 10, lift_GenLLVM (elems_ typs_in_ctx))] ++ ...`

Where `typs_in_ctx` is just the filtered context, not mapped into generators or anything. This way we have an upper bound on how frequently this is used.

Chobbes (changes requested):

src/coq/QC/GenAST.v
hanxic: So, you are saying, for ctx, we need to first pick a specific type in ctx, and then pick one from it along with all other stuffs not in the ctx (the code in later lines). Couldn't it generate even less probability then the current version (maximum of 10), in which I let every one of the type in ctx to have the same frequency (like 1 or 2, so if we have more than 5 things in the ctx it will jump up to 10).

src/coq/QC/GenAST.v
```diff
@@ -530,31 +606,40 @@ Section TypGenerators.
...
-                      ; ret TYPE_Array <*> lift genN <*> gen_sized_typ_size sz'
-                      ; ret TYPE_Vector <*> (n <- lift_GenLLVM genN;;ret (n + 1)%N) <*> gen_sized_typ_size 0
-                      ; let n := Nat.div sz 2 in
-                        ret TYPE_Function <*> gen_typ_size n <*> listOf_LLVM (gen_sized_typ_size n)
-                      ; ret TYPE_Struct <*> nonemptyListOf_LLVM (gen_sized_typ_size sz')
-                      ; ret TYPE_Packed_struct <*> nonemptyListOf_LLVM (gen_sized_typ_size sz')
-                      ]
+    | (S sz') =>
+        ctx <- get_ctx;;
+        let typs_in_ctx := map (fun '(_, typ) => ret typ) (filter_non_void_typs ctx) in
```
Chobbes, line 611: Let's not map `ret` over all of the types in the context. We can just filter the types in the context and then use `elems_` instead of `oneOf` later.

Chobbes (changes requested):

src/coq/QC/GenAST.v
hanxic: elems_ takes in a default value. Should the default value just be the type when size 0 is the input?

src/coq/QC/GenAST.v
Chobbes: I'm not following you.

I don't think we want to make it constantly more likely to generate a type that's already in the context the larger the context is, but the exact weights are something we need to play with #221

hanxic: nvm on that. I made the change

src/coq/QC/GenAST.v
Chobbes: Ahhhhhhhh. Right, okay. We should probably add a version of `elems_` with no default later, but this is fine for now then.

hanxic: ok

Issue #189, opened 2021-12-13
Coq flocq version 4.0.0 breaks IEEE754_extra.v


Chobbes:
I noticed this when getting things set up on another computer. This file complains about Z not existing (I guess flocq no longer exports ZArith), and after importing ZArith it complains about `R` not being a valid scope here:

```
Lemma Bsign_pos0:
  forall x, is_finite_pos0 x = true -> Bsign _ _ x = Rlt_bool (B2R _ _ x) 0%R.
```

Might just be a matter of importing a few things, but this file is also from Compcert, so maybe there's an up-to-date version we should steal?

atrieu:
I'm guessing https://github.com/AbsInt/CompCert/pull/368 would be of interest.

YaZko:
I made the `opam-release` branch compatible with `flocq 4.0`. This will be fixed once things are cleaned up and merged.

saulshanabrook:
I was trying to build `master` and I think hit this bug, so I switched to `opam-release`. 

I now am getting this error when running `make`:

```
patch -p0 < CRelationClasses.mli.patch
can't find file to patch at input line 3
Perhaps you used the wrong -p or --strip option?
The text leading up to this was:
--------------------------
|--- ml/extracted/CRelationClasses.mli	2018-10-08 20:41:02.000000000 -0700
|+++ src/CRelationClasses.mli	2018-10-08 20:41:05.000000000 -0700
--------------------------
File to patch: 
```

I can also open another issue for this, if that is prefered.

YaZko:
Hello @saulshanabrook ,

Ah my apologies: while this particular issue is indeed fix in the `opam-release`, we are currently still fighting in this branch to move from a `make` based build system to a fully `dune` based one: it is not yet ready, hence why you have hit this issue.

I will port `master` to coq `8.15` and `flocq 4.0` on Friday as a quick fix.

Sorry for the trouble and thanks for your interest in Vellvm :)

saulshanabrook:
Thanks for the explanation! That would be great, I'm looking forward to trying out the library on some LLVM examples.

YaZko:
Alright the `dev` branch (now replacing `master`, we had both floating around previously so I took this opportunity to clean up) should now compile against the most recent opam versions of the packages we depend on following PR #193 .  
In particular, it should compile against `coq 8.15`, `flocq 4.0` and `dune 3.0`.

Please note that this breaks backward compatibility, we are officially committing to these versions.

[...]

PR #250, opened 2022-06-13
Pin dune and coq-flocq versions

ConDai:
This PR pins `coq` to version `8.15.0`, `flocq` to version `4.0.0` and `dune`  to version `3.0.3`.

I was having the following compilation issue while following the `README.md` in the branch `8.15`.

```bash
COQC coq/Numeric/IEEE754_extra.v
File "./coq/Numeric/IEEE754_extra.v", line 953, characters 24-33:
Error: The reference BSN.Bplus was not found in the current environment.
```

Looking into the issues I found from https://github.com/vellvm/vellvm/issues/189 that the branch `8.15`  assumes `coq 8.15.0`, `flocq 4.0.0` and `dune 3.0.3`.
 
I believe it would help if this was explicit in the `README.md`. Also it would be convenient if we could install the right versions using `make opam`. Hence, I've updated the instructions to install `flocq` and `dune` in the `README.md` and `make opam` to pin the required versions.

Apologies in advance if this against some best practices. I've never done package management in Coq

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,9 +106,9 @@ Our current test-suite of LLVM programs for which we compare our semantics again
 # Installing / Compiling Vellvm
 
 ### Assumes: 
-  - Coq: version 8.15 (installed via *opam*, see below)
+  - Coq: version 8.15.0 (installed via *opam*, see below)
   - OCaml: version 4.12.0 (installed via *opam*, see below)
-  - opam: version 2.0.8.
+  - opam: version 2.0.8+.
     It is available via [homebrew](https://brew.sh/) on Mac, and most system's package managers on Linux, e.g. `sudo apt-get install opam`.
     If this is the first time you are using opam you need to initialize it: 
     + On Linux: `opam init`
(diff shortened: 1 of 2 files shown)


Chobbes:
Yeah, we should probably pin things like this. Note, in the `dev` branch I believe we're back on `flocq 3.4.3`, I've reverted it to make sure QuickChick still works.

Also for what it's worth... at this point I would just recommend using `nix`, which does pin the versions of everything by default. Once you:

- install `nix`: https://nixos.org/download.html
- enable flakes (to pin versions): https://nixos.wiki/wiki/Flakes

You can just type `nix develop .` in the `vellvm` directory to get a full development environment with the proper version of everything. This should work for all of the recent branches that I've touched. I'd recommend doing this if you intend to look at any different branches, as there are subtle changes in dependencies.

Zdancewic:
The dev branch doesn't compile for me with Coq 8.15.2 and coq-flocq 3.4.3, so I don't know what the current situation is.

Zdancewic:
OK, one other problem was that `coq-itrees` upgrade to v 5.0.0 broke the dev branch too.  I believe that everything is now consistent again with: Coq 8.15.2, coq-flocq 3.4.3, and everything else the default opam release version (including coq-itrees 5.0.0).  I also updated the README with these version numbers.

ConDai:
Thanks @Chobbes! I'll look into using nix.

ConDai:
Hi @Zdancewic. That was the problem I was facing.  Thanks for the update!

PR #255, opened 2022-06-17
Bug fixed for insert value

### LINKED ISSUE #254: gen_insertvalue concerns with generating a value to be inserted
`gen_insertvalue` creates a list of paths to fields within an aggregate type. These fields may be pointer types, or may be aggregate types that contain pointer types... And we then try to generate an expression of size 0 with the type of this field. Unfortunately this will fail in most circumstances, unless there exists a pointer of the appropriate type in the context already.

```
Definition gen_insertvalue : GenLLVM (typ * instr typ) :=
  '(id, tagg) <- get_ctx_agg_typ;;
  let paths_in_agg := get_index_paths_agg tagg in
  '(tsub, path_for_insertvalue) <- oneOf_LLVM (map ret paths_in_agg);;
  ex <- hide_ctx (gen_exp_size 0 tsub);;
  (* Generate all of the type*)
  ret (tagg, INSTR_Op (OP_InsertValue (tagg, EXP_Ident id) (tsub, ex) path_for_insertvalue)).
```

- [ ] We should make sure that we only include paths with pointers that are already in the context
- [ ] We should make sure that `gen_insertvalue` is only called when it is possible to generate an `insertvalue` instruction. I.e., if all possible paths contain pointers that we don't have in the context, then `gen_insertvalue` should be skipped in `gen_instr`.

hanxic:
^. Now it should only detect whether the aggregate type has valid path and will only select from the paths that work.

### DIFF
--- a/src/coq/QC/GenAST.v
+++ b/src/coq/QC/GenAST.v
@@ -1662,6 +1662,7 @@ Section InstrGenerators.
   Definition gen_definition (name : global_id) (ret_t : typ) (args : list (local_id * typ)) : GenLLVM (definition typ (block typ * list (block typ)))
     :=
       ctx <- get_ctx;;
+      ptr_int_ctx <- get_ptrtoint_ctx;;
       (* Add arguments to context *)
       let args_ctx := map (fun '(i, t) => (ID_Local i, t)) args in
       append_to_ctx args_ctx;;
@@ -1679,6 +1680,7 @@ Section InstrGenerators.
       in
       (* Reset context *)
       modify (replace_ctx ((ID_Global name, f_type) :: ctx));;
+      modify (replace_ptrtoint_ctx ptr_int_ctx);;
       ret (mk_definition (block typ * list (block typ)) prototype (map fst args) bs).
 
   Definition gen_new_definition (ret_t : typ) (args : list (local_id * typ)) : GenLLVM (definition typ (block typ * list (block typ)))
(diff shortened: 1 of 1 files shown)


src/coq/QC/GenAST.v
```diff
@@ -1296,9 +1343,10 @@ Definition genType: G (typ) :=
           end in
       let fix gen_size_0 (t: typ) :=
           match t with
-          | TYPE_I n                  => ret EXP_Integer <*> lift (arbitrary : G Z) (* lift (x <- (arbitrary : G nat);; ret (Z.of_nat x)) (* TODO: should the integer be forced to be in bounds? *) *)
+          | TYPE_I n                  => ret EXP_Integer <*> lift (choose (0, (BinIntDef.Z.of_N n) - 1)) (* lift (x <- (arbitrary : G nat);; ret (Z.of_nat x)) (* TODO: should the integer be forced to be in bounds? *) *)
```
Chobbes: What is this about? It looks like you'll just pick between 0 and 7 for an `i8` when it should be 0 and 256, right?

Also, is this change part of this PR? It seems like it should be in a separate small one (it's easier to manage commits if it's sort of 1 feature per PR).

src/coq/QC/GenAST.v
```diff
@@ -510,7 +513,7 @@ Definition filter_ptr_vecptr_typ (ctx: list (ident * typ)) : list (ident * typ)
             | TYPE_Pointer _ => true
             | TYPE_Vector _ (TYPE_Pointer _) => true
             | _ => false
-            end) ctx.
+            end) ctx.
```
Chobbes: Some extra whitespace in a few places `M-x delete-trailing-whitespace` will get rid of it.

Chobbes (changes requested):
A few initial comments, I need to look in more detail for the actual meat of the PR, though!

src/coq/QC/GenAST.v
hanxic: Sure I will revert this one

PR #256, opened 2022-06-20
gen_exp_size better integer

hanxic:
Generate integer that are the minimum of its size and (2^30 - 1). The reason we have 2^30 - 1 and not the regular 2^64 - 1 is because in `QuickChick/src/ExtractionQC.v.cppo`, given our version is below 4.13.0, the ocaml random integer function QC is calling is `Random.int`, which allows a range of [0,2^30), instead of `Random.full_int`, which allows a random integer beyond the range.

Another thing that we can probably decide is whether we want to generate negative integer. Currently we only generate positive integer. The PR so far will make sure integers are in bound. Don't know how this will affect some of the tests tho so maybe need further guidance on this.

### DIFF
--- a/src/coq/QC/GenAST.v
+++ b/src/coq/QC/GenAST.v
@@ -472,45 +472,45 @@ Section GenerationState.
 End GenerationState.
 
 Section TypGenerators.
-    (*filter all the (ident, typ) in ctx such that typ is a ptr*)
-Definition filter_ptr_typs (ctx : list (ident * typ)) : list (ident * typ) :=
-  filter (fun '(_, t) => match t with
+  (*filter all the (ident, typ) in ctx such that typ is a ptr*)
+  Definition filter_ptr_typs (ctx : list (ident * typ)) : list (ident * typ) :=
+    filter (fun '(_, t) => match t with
                         | TYPE_Pointer _ => true
                         | _ => false
-                     end) ctx.
+                        end) ctx.
 
   Definition filter_sized_typs (typ_ctx: list (ident * typ)) (ctx : list (ident * typ)) : list (ident * typ) :=
     filter (fun '(_, t) => is_sized_type typ_ctx t) ctx.
 
-Definition filter_non_void_typs (ctx : list (ident * typ)) : list (ident * typ) :=
-  filter (fun '(_, t) => match t with
-                      | TYPE_Void => false
-                      | _ => true
-                      end) ctx.
+  Definition filter_non_void_typs (ctx : list (ident * typ)) : list (ident * typ) :=
+    filter (fun '(_, t) => match t with
+                        | TYPE_Void => false
+                        | _ => true
+                        end) ctx.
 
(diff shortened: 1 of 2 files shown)


Chobbes:
There's a lot of commits in here and I'm not really sure what's going on. If you want you could look up how to do `git rebase` / `git squash` and clean this up (note: rewriting history can be dangerous, so make a copy of the branch just in case!). This PR should probably just have the one commit, and should only change `GenAST.v`, but it changes some QC files too, which seems odd.

Alternatively, you can use `git reset` on the branch (this will lose work, again make a copy of the branch just in case!) to make it the same as `memory-testing` itself, and then you can make the changes to `GenAST.v` again.

Regardless, you will have to force-push to this branch!

src/coq/QC/GenAST.v
```diff
@@ -1296,7 +1296,12 @@ Definition genType: G (typ) :=
           end in
       let fix gen_size_0 (t: typ) :=
           match t with
-          | TYPE_I n                  => ret EXP_Integer <*> lift (arbitrary : G Z) (* lift (x <- (arbitrary : G nat);; ret (Z.of_nat x)) (* TODO: should the integer be forced to be in bounds? *) *)
+          | TYPE_I n                  =>
+              let size := BinIntDef.Z.of_N (N.min (2 ^ n) (2 ^ 30 - 1)) in
```
Chobbes: You are probably right about the ocaml integers issue, and we should look into that (I think these may get extracted to big ints now, but I guess you ran into problems that suggest otherwise?)

Regardless, we should make sure it works for signed values...

And ideally we wouldn't have this `2^30 - 1` bound, because that would mean we couldn't generate things like `2^32 - 1`, which should be signed `-1` for `i32` values... That seems like a useful number to be able to generate. Furthermore, with `i64` signed values you wouldn't be able to generate any negative numbers :(.

src/coq/QC/GenAST.v
```diff
@@ -1200,7 +1200,7 @@ Fixpoint gen_typ_le_size (max_byte_sz : N) : GenLLVM typ :=
   oneOf_LLVM
     ( (* Primitive types *)
       get_prim_typ_le_size max_byte_sz ++
-
+
```
Chobbes: Whitespace.

src/coq/QC/GenAST.v
```diff
@@ -267,7 +267,7 @@ Section GenerationState.
         ; gen_typ_ctx := gs.(gen_typ_ctx)
         ; gen_ptrtoint_ctx := ptrtoint_ctx
        |}.
-
+
```
Chobbes: Whitespace.

Chobbes (changes requested):

src/coq/QC/GenAST.v
hanxic: Big ints can work if we are able to switch to OCaml version of 4.13.0, which I don't think we are.
What I think I can do, for i64 now, is to generate another 2^32 - 1, and then multiply them together

PR #264, opened 2022-06-24
Solution for QC stuck issue

### LINKED ISSUE #221: QuickChick Performance

hanxic:
Solution for issue #221. Solve by bounding the shift operations. Is generating failure cases irrelevant to shift (current record 4 run: 1 success, 3 failure, 0 stuck). Also change the sequence of testing by first going through clang and then go through the interpreter

### DIFF
--- a/src/coq/QC/GenAST.v
+++ b/src/coq/QC/GenAST.v
@@ -50,8 +50,6 @@ Open Scope Z_scope.
     not used in proofs) it's not terribly important to prove that they
     actually terminate.  *)
 Unset Guard Checking.
-
-(* TODO: Move this? *)
 (** Difference lists *)
 Section DList.
   Definition DList (A : Type) := list A -> list A.
@@ -688,7 +686,6 @@ Definition filter_ptr_vecptr_typ (ctx: list (ident * typ)) : list (ident * typ) 
                 (* ; TYPE_X86_mmx *)
                 (* ; TYPE_Opaque *)
                 ])).
-
   Program Fixpoint gen_typ_non_void_size (sz : nat) {measure sz} : GenLLVM typ :=
     match sz with
     | 0%nat => gen_typ_non_void_0
@@ -767,8 +764,25 @@ Definition filter_ptr_vecptr_typ (ctx: list (ident * typ)) : list (ident * typ) 
 End TypGenerators.
 
 Section ExpGenerators.
-  (* nuw / nsw make poison values likely *)
+
   Definition gen_ibinop : G ibinop :=
+     oneOf_ failGen
+           [ ret LLVMAst.Add <*> ret false <*> ret false
+           ; ret Sub <*> ret false <*> ret false
+           ; ret Mul <*> ret false <*> ret false
+           ; ret Shl <*> ret false <*> ret false
+           ; ret UDiv <*> ret false
+           ; ret SDiv <*> ret false
+           ; ret LShr <*> ret false
+           ; ret AShr <*> ret false
+           ; ret URem
+           ; ret SRem
+           ; ret And
+           ; ret Or
+           ; ret Xor
+           ].
+  (*
+  Definition gen_ibinop (t: typ) (ex1 ex2: exp typ) : G ibinop :=
     (* Note: some of these binops are currently commented out due to a
        bug with extraction and QC. *)
     oneOf_ failGen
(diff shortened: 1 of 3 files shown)


src/coq/QC/GenAST.v
```diff
@@ -50,7 +50,14 @@ Open Scope Z_scope.
     not used in proofs) it's not terribly important to prove that they
     actually terminate.  *)
 Unset Guard Checking.
-
+Axiom print_extracted_coq_string : string -> unit.
```
Chobbes: Clean up.

src/coq/QC/GenAST.v
```diff
@@ -768,24 +775,49 @@ End TypGenerators.
 
 Section ExpGenerators.
   (* nuw / nsw make poison values likely *)
-  Definition gen_ibinop : G ibinop :=
+
+  Definition get_bitwidth (z: Z) : Z :=
```
Chobbes: Why do we need this? Don't we already have the bitwidths?

src/coq/QC/GenAST.v
```diff
@@ -768,24 +775,49 @@ End TypGenerators.
...
+    Z.log2 z + 1.
+  
+  Definition gen_ibinop (t: typ) (ex1 ex2: exp typ) : G ibinop :=
```
Chobbes: Why change `gen_ibinop`? I think it should stay the same.

src/coq/QC/GenAST.v
Chobbes: Oh, I think you're using this for the shift amount, but you don't have to. If you have an i32 you can only shift between 0 and 31 bits. No need to take the log of anything.

src/coq/QC/GenAST.v
```diff
@@ -768,24 +775,49 @@ End TypGenerators.
...
     (* Note: some of these binops are currently commented out due to a
        bug with extraction and QC. *)
+    let z1 :=
```
Chobbes: These matches are a bit odd, but we should be able to get rid of them and put `gen_ibinop` back how it was.

src/coq/QC/GenAST.v
Chobbes: I think these changes should just be made in `gen_ibinop_exp_typ`.

src/coq/QC/GenAST.v
```diff
@@ -1658,9 +1698,11 @@ Section InstrGenerators.
          (bound : LLVMAst.int) {struct t} : GenLLVM (terminator typ * (block typ * list (block typ)))
     :=
       bid_entry <- new_block_id;;
+      (*let _ := print_extracted_coq_string "tESTing" in*)
```
Chobbes: Cleanup.

src/coq/QC/GenAST.v
```diff
@@ -1658,9 +1698,11 @@ Section InstrGenerators.
...
       (* TODO: make it so I can generate constant expressions *)
       loop_init <- ret INSTR_Op <*> gen_op (TYPE_I 32 (* TODO: big ints *));; (* gen_exp_size sz (TYPE_I 64);; *)
       bound' <- lift_GenLLVM (choose (0, bound));;
+      (* let bound' := 0%Z in *)
```
Chobbes: Cleanup.

Chobbes (changes requested):

Chobbes:
Where are the changes to `lib/` coming from?

hanxic:
I didn't touch it at all. This one is directly copied form memory-testing....

src/coq/QC/GenAST.v
hanxic: Ok, but if we have like i8 4, wouldn't shifting it 7 bits to the left makes it overflow? Or will it just shift to negative

Issue #209, opened 2022-05-26
Generate Function calls


Chobbes:
Add function calls to generated LLVM programs. First step is to look at how `gen_main` works, and then thinking about how we want to generate function calls while avoiding recursive calls. We may need to modify `GenState` so that the generators have information about which functions can be called in a program.

YaZko:
GenLLVM should be pulled out as a separate project, and this issue migrated there.

PR #275, opened 2022-07-18
Generate functions mentioned in Issue #209

hanxic:
For testing purposes, can only generate 0, 1, or 2 functions. Can edit function `gen_helper_function_tle_multiple` to generate a wider range number of functions. Currently, the solution for function recursion is that the function cannot call itself but only functions created before it. Will update to include recursion call

### DIFF
--- a/src/coq/QC/GenAST.v
+++ b/src/coq/QC/GenAST.v
@@ -1996,11 +1996,15 @@ Section InstrGenerators.
        | Raw _ => false
        end.
   (* Don't want to generate CFGs, actually. Want to generated TLEs *)
-  Definition gen_definition (name : global_id) (ret_t : typ) (args : list (local_id * typ)) : GenLLVM (definition typ (block typ * list (block typ)))
+
+  Definition gen_definition (name : global_id) (ret_t : typ) (args : list (typ)) : GenLLVM (definition typ (block typ * list (block typ)))
     :=
       ctxs <- get_variable_ctxs;;
 
       (* Add arguments to context *)
+      args <- map_monad (fun t =>
+                        i <- new_raw_id;;
+                        ret (i, t)) args;;
       let args_ctx := map (fun '(i, t) => (ID_Local i, t)) args in
       append_to_ctx args_ctx;;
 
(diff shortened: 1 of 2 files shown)


hanxic:
Actually recursive call is said to be avoided in this issue. I think this is the solution then

Chobbes:
I'm looking into this PR but I'm getting a lot of issues. I have things rebased locally with some changes (209-calvin-fixes branch) and I can force push if you don't mind (you'll have to reset your branch against mine if I do that).

I don't understand why yet, but I'm getting a lot of discards after generating functions. It seems like there's some weirdness with function types being kind of implicitly cast to pointers with the `typ_to_dtyp` conversion, which maybe messes up the `is_sized_typ` check (I'm not 100% sure how it should be treated yet, to be honest). You've worked around this issue in `gen_sized_typ_size_ptrinctx` by filtering out function types first, but if this needs to be done here, I'd guess it needs to be done in `gen_sized_type_size` as well? I'm not 100% sure on the details yet.

I'm wondering if perhaps when you generate a function `@g1` the type should be stored in the context as a pointer to a function, and not just the function type itself. This would match up with how these identifiers are actually used... HMMM.

Anyway, I'll look into making some of these changes after lunch.

hanxic:
I

> I'm looking into this PR but I'm getting a lot of issues. I have things rebased locally with some changes (209-calvin-fixes branch) and I can force push if you don't mind (you'll have to reset your branch against mine if I do that).
> 
> I don't understand why yet, but I'm getting a lot of discards after generating functions. It seems like there's some weirdness with function types being kind of implicitly cast to pointers with the `typ_to_dtyp` conversion, which maybe messes up the `is_sized_typ` check (I'm not 100% sure how it should be treated yet, to be honest). You've worked around this issue in `gen_sized_typ_size_ptrinctx` by filtering out function types first, but if this needs to be done here, I'd guess it needs to be done in `gen_sized_type_size` as well? I'm not 100% sure on the details yet.
> 
> I'm wondering if perhaps when you generate a function `@g1` the type should be stored in the context as a pointer to a function, and not just the function type itself. This would match up with how these identifiers are actually used... HMMM.
> 
> Anyway, I'll look into making some of these changes after lunch.

I'm totally fine with rebase... I don't know what I need to do but just let me know.

Issue #284, opened 2022-07-25
dvalue / uvalue extraction build issue


Chobbes:
For some reason in `MemoryModelImplementation.mli` won't build by default. It produces this (same issue with `uvalue` too):

```
        type dvalue = LLVMParamsBigIntptr.Events.DV.dvalue =
        | DVALUE_Addr of ADDR.addr
        | DVALUE_I1 of int1
        | DVALUE_I8 of int8
        | DVALUE_I32 of int32
        | DVALUE_I64 of int64
        | DVALUE_IPTR of IP.intptr
        | DVALUE_Double of ll_double
        | DVALUE_Float of ll_float
        | DVALUE_Poison of dtyp
        | DVALUE_None
        | DVALUE_Struct of dvalue list
        | DVALUE_Packed_struct of dvalue list
        | DVALUE_Array of dvalue list
        | DVALUE_Vector of dvalue list
```

Changing it to this makes the build succeed.

```
        type dvalue = LLVMParamsBigIntptr.Events.DV.dvalue =
        | DVALUE_Addr of LLVMParamsBigIntptr.ADDR.addr
        | DVALUE_I1 of int1
        | DVALUE_I8 of int8
        | DVALUE_I32 of int32
        | DVALUE_I64 of int64
        | DVALUE_IPTR of LLVMParamsBigIntptr.IP.intptr
        | DVALUE_Double of ll_double
        | DVALUE_Float of ll_float
        | DVALUE_Poison of dtyp
        | DVALUE_None
        | DVALUE_Struct of dvalue list
        | DVALUE_Packed_struct of dvalue list
        | DVALUE_Array of dvalue list
        | DVALUE_Vector of dvalue list
```

Chobbes:
Addressed in 2d65026a

PR #299, opened 2022-08-05
More parsing

Zdancewic:
This seems correct, but generates this type error upon extraction:

```
File "ml/extracted/MemoryModelImplementation.mli", lines 5291-5305, characters 8-38:
5291 | ........type dvalue = LLVMParamsBigIntptr.Events.DV.dvalue =
5292 |         | DVALUE_Addr of ADDR.addr
5293 |         | DVALUE_I1 of int1
5294 |         | DVALUE_I8 of int8
5295 |         | DVALUE_I32 of int32
...
5302 |         | DVALUE_Struct of dvalue list
5303 |         | DVALUE_Packed_struct of dvalue list
5304 |         | DVALUE_Array of dvalue list
5305 |         | DVALUE_Vector of dvalue list
Error: This variant or record definition does not match that of type
         LLVMParamsBigIntptr.Events.DV.dvalue
       Constructors do not match:
         DVALUE_Addr of LLVMParamsBigIntptr.ADDR.addr
       is not compatible with:
         DVALUE_Addr of ADDR.addr
       The types are not equal.
make[1]: *** [_build/default/ml/main.exe] Error 1
make: *** [all] Error 2
```


Chobbes:
Yeah, this is the same as #284 I'm not really sure what's happening. I've been patching it manually, but I guess we should add it to the `fix-extraction.sh` script.

Zdancewic:
I think it's OK to merge this now.

PR #300, opened 2022-08-07
Fix the bug on bitcast vector generator

### LINKED ISSUE #301: Bitcast to correct type bug
Sometimes we encounter situations like 
`temporary_vellvm.ll:14:19: error: invalid cast opcode for cast from '<3 x float>' to '<1 x double>'
    %v5 = bitcast <3 x float> %v4 to <1 x double>`
Need to make sure that the type is more accurate

hanxic:
Fixing bitcast generator, before it may generate code like `%v8 = bitcast <1 x i1> %v4 to <0 x i8>`. Clang will complain about this. Essentially we need to not only check the division but also the mod.


hanxic:
See #305

PR #304, opened 2022-08-21
Fix for gen exp for function type

hanxic:
Previously, it is allowed for something like `alloca ... @g4 ...` where `@g4` is a function to generate (apologize for not being able to give an example since I lost track....). However, I don't think we should allow that. The reason why this is happening is because we store function as a pointer in the context. Maybe it's worth consider to only store them as function (and then whenever we want to call we add a `TYPE_Pointer` around the `TYPE_Function` that we want to call from).


hanxic:
See #305

PR #339, opened 2022-11-07
`raise_oom`, `raise_error`, `raise_ub` absurd inversions

euisuny:
Discharging obligations of the kind `~(raise_oom m \approx raise_ub)` (and friends)

Had to change the event signatures to leave all the events concrete except for the call events (`CallE`, `ExternalCallE`), since these obligations were not provable when leaving the event signatures totally abstract.


Chobbes:
Cool, looks good to me, I'll merge this as soon as github confirms that it builds :).

PR #344, opened 2023-02-12
Support the Coq Platform by explicitly using the coq-flocq3 package.

paul-snively:
Thanks for all the hard work on vellvm!

I'm trying to use the [Coq Platform]() for a variety of tasks, and as of this writing, vellvm makes that slightly awkward because it depends on Flocq version 3, but expects to find it at the logical path `Flocq`, which, in the Coq Platform, refers to Flocq version 4. Luckily, the Coq Platform maintainers provide distinct virtual packages, coq-flocq3 and coq-flocq4, for this reason. This Pull Request simply changes vellvm to depend on coq-flocq3 and use the `Flocq3` logical path instead. The system compiles and all of the tests but one pass, which I will report an issue about.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -6,7 +6,7 @@ COQLIBDIR = ../lib
 MLDIR = ml
 EXTRACTDIR = ml/extracted
 
-OPAMPKGS=coq coq-ext-lib coq-paco coq-ceres coq-flocq coq-mathcomp-ssreflect coq-simple-io coq-itree cppo dune menhir qcheck ocamlbuild
+OPAMPKGS=coq coq-ext-lib coq-paco coq-ceres coq-flocq3 coq-mathcomp-ssreflect coq-simple-io coq-itree cppo dune menhir qcheck ocamlbuild
 
 QUICKCHICKDIR=../lib/QuickChick
 FLOCQQUICKCHICKDIR=../lib/flocq-quickchick
--- a/src/coq/Numeric/Archi.v
+++ b/src/coq/Numeric/Archi.v
@@ -18,7 +18,7 @@
 (** Architecture-dependent parameters for PowerPC *)
 
 Require Import ZArith List.
-From Flocq Require Import Binary Bits.
+From Flocq3 Require Import Binary Bits.
 
 Definition ptr64 := false.
 
--- a/src/coq/Numeric/Floats.v
+++ b/src/coq/Numeric/Floats.v
@@ -20,7 +20,7 @@
 Require Import Coqlib Zbits Integers.
 Require Import Coq.micromega.Lia.
 (*From Flocq*)
-From Flocq Require Import Binary Bits Core.
+From Flocq3 Require Import Binary Bits Core.
 Require Import IEEE754_extra.
 Require Import Program.
 Require Archi.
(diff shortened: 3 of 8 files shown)


Chobbes:
I haven't used the Coq Platform before, but it'd be great if we could make sure that Vellvm works with it! We noticed that `coq-flocq3` is a thing that only exists on the Coq Platform, so this wouldn't work with people just using plain old opam, right? Is there a way to make it work for both?

Zdancewic:
Thanks for the interest @paul-snively!  It seems that this change does not work unless the development is being built in the context of Coq Platform.   Is there a way to support Coq Platform and non-Coq Platform development?  Otherwise, we'll have to give some thought about how to best support it.

paul-snively:
> Thanks for the interest @paul-snively! It seems that this change does not work unless the development is being built in the context of Coq Platform. Is there a way to support Coq Platform and non-Coq Platform development? Otherwise, we'll have to give some thought about how to best support it.

That's a really good question. I _think_ we can (but haven't yet tried to) pass an `-R` argument to `coqc` via the command line to remap the logical path to `Flocq3` to refer to the non-Coq-Platform `coq-flocq` module, pinned as you have documented. Since `coq-flocq3` from the Coq Platform is just a virtual package, I'm hoping that the `-R` argument at the command-line would effectively be redundant in that case ("remap `Flocq3` to `Flocq3`"), but would be effective in the non-Coq-Platform case.

Thoughts?

Incidentally, thank all of you for your brilliant work on "Software Foundations," Vellvm, and Coq generally. I'm very much looking forward to digging into it in more depth. In particular, I'm interested in formalizing and proving various things about the [Verse Calculus](https://simon.peytonjones.org/assets/pdfs/verse-conf.pdf), and any comments any of you may have about, e.g. proving some of the traditional things about lambda calculi or extensions, such as confluence, would be greatly appreciated. Ultimately, I'd love to use Vellvm to write a simple compiler, perhaps with a REPL for some surface syntax, for the Verse Calculus, and essentially try to track where the Epic team goes with the calculus, as well as, eventually, with their own full-blown language.

One last thing: I see that the test failure I'm experiencing is expected. Is there some sort of roadmap for the various branches the project has somewhere? It seems `dev` isn't the appropriate one to develop from at the moment, but it's not quite clear to me what is. :-)

Thanks again!

Chobbes:
> One last thing: I see that the test failure I'm experiencing is expected. Is there some sort of roadmap for the various branches the project has somewhere? It seems dev isn't the appropriate one to develop from at the moment, but it's not quite clear to me what is. :-)

There's some pretty big changes coming down the line in the `finite-memory` branch, but it's not really ready for prime time just yet! It might be a little while still. This probably makes `dev` the most appropriate one to develop against for now, and you wouldn't be alone as that's essentially the version that the [Helix project](https://github.com/vzaliva/helix) is developing against. The `finite-memory` branch once merged will likely not be a simple drop-in replacement, as we've changed a lot about how memory, undef, and undefined behaviour works. I'm hoping that if you don't depend on a lot of weird edge cases in these things that we'll be able to provide some lemmas to bridge the gap somewhat, but it's a pretty big change to vellvm, so I'm not sure how well that will pan out.

Zdancewic:
I'm closing this pull request, as it seems that the [rocq-platform](https://rocq-prover.org/platform) now supports coq-flocq 4.2.0, which should be compatible.  

It would still be good to confirm that the new Vellvm `main` branch compiles cleanly with `rocq-platform`, but we should revisit that as separate issues / pull requests.

PR #347, opened 2023-03-02
Remove quickchick lib

Zdancewic:
This pull requests removes the dependency on `lib/QuickChick` and should compile with:
```
coq                    8.15.2      pinned to version 8.15.2
coq-ceres              0.4.0       Library for serialization to S-expressions
coq-ext-lib            0.11.7      A library of Coq definitions, theorems, and tactics
coq-flocq              4.1.1       A formalization of floating-point arithmetic for the Coq system
coq-itree              5.1.0       Library for representing recursive and impure programs with equational reasoning
coq-mathcomp-ssreflect 1.16.0      Small Scale Reflection
coq-paco               4.1.2       Coq library implementing parameterized coinduction
coq-quickchick         1.6.4       Randomized Property-Based Testing Plugin for Coq
coq-simple-io          1.8.0       IO monad for Coq

```


Zdancewic:
For now, I have added an axiom for `failGen`, but that does seem to trigger an error when we do `make qc-test`.

Zdancewic:
@Chobbes - I took a pass through the quickchick generator to get rid of `failGen` by making GenLLVM have an option inside it.  The result is that `make qc-tests` seems to run (and spew some output) without triggering the axiom.
To achieve this I
- commented out a couple of functions that don't seem to be used, and
- in a few places chose an arbitrary element to pass to `oneOf_` as (what I think is) the default in case the list is empty

Let me know what you think about this.

PR #356, opened 2024-03-29
Changing the representation of cfgs and other changes

YaZko:
# Motivation and context

_Note_: Currently built atop of the `fast-install` branch, i.e. the latest icfp submission branch with the adequacy between finite and infinite levels commented out.

This ongoing PR has essentially two goals:
1. To move from cfgs represented as a list of blocks containing an identifier to finite maps from identifiers to "anonymous" blocks.
2. Since the change forces me to go through the whole code base, to take the opportunity for other changes, and for taking stock of future changes.

The overall motivation for 1. is straightforward: the list representation had virtually no good property:
1. It may contain several blocks with the same id, forcing us to carry the `wf_ocfg_bid` everywhere we would want to manipulate the graphs
2. It is algorithmically quite terrible obviously
3. It has not much dedicated theory, and no associated automation (forcing me to half-hack things like `solve_find_block`

I believe that the main reason we bear with it so far is that we have done surprisingly little in terms of AST manipulation so far, except for a focused effort leading to `ScopeTheory.v` when we proved the block fusion optimization, and sone around `solve_find_block` and the like during the Helix project.
However, we are at the moment working on graph (de)construction for Vellvm with @lfrenot and @drup, and the need for a better representation feels quite necessary.

There are as far as I know two representations of finmap that stand out for ticking out all the right boxes in their design: Leroy and Appel original's extensional binary tries¹, and the stdpp's gmap implementation that it inspired. We favour the latter for their extensive theory, type-class-based approach, and exceptional automation.

# Status

Still working mostly through the syntax-related files for now, albeit it is of course where the most work is expected.

# Main changes

## Representation of graphs

Blocks lose their `blk_id` field, but the notion of `blockID` is introduced for the places where it remains useful (for now, at the interface with the top level entities). Open graphs are now finmaps from `block_id` to `block`.

```coq
Record block : Set :=
  mk_block
    {
      (* blk_id    : block_id; *)
      blk_phis  : list (local_id * phi);
      blk_code  : code;
      blk_term  : terminator;
      blk_comments : option (list string)
    }.

Definition blockID := (block_id * block)%type.

(* A notation so that we don't have to list stdpp's instances *)
Notation ocfg := (gmap block_id (block T)).
```

## stdpp's classes for countability of sets and decidability of equality

The interface for finmaps in stdpp is provided by a rich family of classes. `gmap` is an efficient implementation provided the keys have decidable equality and are countable. We hence move to these classes instead of the previously used decidable equality from extlib.

```coq
#[global] Instance EqDecision_raw_id : EqDecision raw_id.
#[global] Instance Countable_raw_id : Countable raw_id := (...)
#[global] Instance EqDecision_block_id : EqDecision block_id.
#[global] Instance EqDecision_instr_id : EqDecision instr_id.
#[global] Instance EqDecision_ident : EqDecision ident.
```

## Graph invariants (Scope and ScopeTheory)

Naturally the most impacted part of the development (so far) are the predicate and theory about graphs.
Definitions for inputs and outputs remain semantically the same, but with saner definitions. In particular, they are now sets of ids (`gset block_id`) rather than lists, that could once again contain duplicates.
The `wf_ocfg_bid` predicate ensuring that no two blocks have the same id disappears, being an invariant of the structure.

```coq
  Definition inputs (ocfg : @ocfg T) := dom ocfg.
  Definition successors (bk : block T) : gset block_id :=
    terminator_outputs (blk_term bk).
  Definition outputs_acc : block_id -> block T -> gset block_id -> gset block_id :=
  fun _ bk acc => acc ∪ successors bk.
  Definition outputs (bks : ocfg T) : gset block_id :=
  map_fold outputs_acc ∅ bks.
  (* Probably useless to cast into a boolean *)
  Definition is_predecessor (b : block_id) (bk : block T) : bool :=
    if decide (elem_of b (successors bk)) then true else false.
  Definition predecessors (b : block_id) (bks : ocfg T) : list block_id :=
    map_fold (fun k bk acc => if is_predecessor b bk then k :: acc else acc) [] bks.

  Definition no_reentrance (bks1 bks2 : ocfg T) : Prop :=
    outputs bks2 ## inputs bks1.
  Definition disjoint_bid (bks1 bks2 : ocfg T) : Prop :=
    inputs bks1 ## inputs bks2.
  Definition free_in_cfg (g : ocfg T ) (id : block_id) : Prop :=
    id ∉ dom g.
```

Porting `ScopeTheory.v` is a bit of work, but essentially because I am only getting familiar with `stdpp`. Proofs like the following are extremely satisfying, and argue in favor of this overall change being worth it.

```coq
  Lemma no_reentrance_not_in (g1 g2 : ocfg T) :
    no_reentrance g1 g2 ->
    forall i, i ∈ outputs g2 -> i ∉ inputs g1.
  Proof.
    set_solver.
  Qed.
```

# Orthogonal changes

## Cutting off Ceres

I implemented years ago show methods for both the internal and ir representation. These implementations relied on serialization into sexpr via the Ceres library. Since then, completely independent and more complete show instances have been implemented, but for some reason the old stuff has been left. I'm cutting it off from the code base.

## Cutting off TypeUtils.v

This file contain unused code, related to the type system, inherited from the original version of Vellvm. We had at some point said we might as well keep it around as it could be useful some day. I believe cleaning up the code base to be more important Today, and don't know if anybody still knows what the content of this file was for. We can always revive it if it becomes relevant some day.

## Removing the specialized signature for expressions

We currently have a specialized signature for expressions compared to say instructions. While it is pretty on paper, it forces us to fight with translations, for little benefits. Let's cut it off.

## Renaming

- [ ] `block_id` -> `bid` --- Not sure, to decide

## Avoiding matching on integers

There are a lot of patterns of the shape `match sz with | 1 | 8 | 16 | ... | _ ...` which are a disaster to work with. Replace all of them with `match suppported sz with | S1 | S8 | ...` with `supported` a finite type?

- [ ]  TODO

# Questions raised on the way

## Representation of top-level entities

Top-level entities, the structures output by the parser before being translated into proper programs, still represent function definitions as plain lists. Is that fine as is, or should it change?

At the moment, I have commented out the proof `modul_of_mcfg_of_modul`: reestablishing it would require a well-formedness invariant on the input module. Do we want it?

## On the functorialization of the syntax

The trick of turning the whole syntax into a functor (in the Haskell sense) so that we factorize the code for exactly two instances of the functor, typ and dtyp, is elegant, but it makes some things a bit clumsy. Part of me is wondering if it’s worth it at all, but at least I’m trying to specialize as much things as possible.

Questions:
- should static analyses or program transformations be run on the typ version, or  be on the dtyp one?
- should the «haskell functor» be a «ocaml functor»? I can’t believe I’m the one saying that, but so that we get our two specializations out of the box right after instead of generalizing functions to functors for no good reasons all over the place, and feed explicit arguments to every piece of syntax?
- or alternatively should there be a Typ typeclass whose typ and dtyp are instances?

The problem with the current situation notably shows:
- in the clutter of every type of syntactic components all across the code base with the notion of type considered
- the awkwardness of `Traversal.v` to provide instances traversing functors rather than plain syntax
- more importantly, type class resolution gets sometime tripped by it. For instance in this PR, because of the parametrization, a `ocfg T` becomes a notation for `(fun T => gmap block_id (block T))` T which requires to provide explicit additional arguments in many places to help the type class inference undertand that it should actually look for instances over `gmap block_id (block T)`.

## More gsets and gmaps?

We use lists and association lists in lieu of sets and finmaps in other places. Should we replace all of them?

todo: collect instances and decide for each of them

- [ ] global maps
- [ ] local maps
- [x] phi nodes (mapping block id to expressions)
- [ ] list of phi nodes in a block (mapping raw_id to phi nodes)
- [ ] list of instructions in a block (mapping instr_id to instructions)

## Cutting off ExtLib

While ExtLib is a great library, it is redundant with stdpp while lacking the modern design, extensive support, and more importantly user base. I believe this should be the opportunity to progressively replace all data structures inherited from Extlib by stdpp ones.

## Adapting to stdpp notations

The library already has notations for bind, which intersect with those from ExtLib but are sensibly different, we would need to adapt to them.

-------------------------------------
¹: https://arxiv.org/abs/2110.05063


Chobbes:
@YaZko a while ago I asked if I could flip the direction of this PR (it's trying to merge dev into gmap-wip and not the other way around), but unfortunately I found out that there isn't actually a way to reverse the direction of a PR on Github. If there's no objection, I think we should close this one and reopen it merging into dev instead of the flipped way. Then it won't run CI and yell at you all of the time :).

I'm happy to do it, but if you want to open the new PR so you get subscribed to notifications for it or whatever that might make the most sense?

YaZko:
Sure please go ahead and close/reopen as fit!

PR #368, opened 2024-07-02
Parseable printf.ll with examples

rogerburtonpatel:
## Main changes: 

Added `va_arg` and `fneg` to the parser. `fneg e` desugars to `fsub double 0x0 e`. 

Added `utilities/vaargs`, which contains printf implementations - hence the large number of files changed. Currently, the two target examples: `utilities/vaargs/dev/working/complex/printf.ll` and 
`utilities/vaargs/dev/working/simpler/printf.ll`-- are both runnable with the parse-updated `vellvm` with no issue. 

## Main review request: 

Is it appropriate to add the `KW_FNEG` to both `instr_op` and `expr_op` in `llvm_parser.mly`? 
Thanks.


Chobbes:
Do you know which `printf` we're probably going to use? Might be good to pair this down to just one implementation so we don't get confused about which version is actually used in the future.

We should probably get rid of the `nanoprintf.h.gch` binary file too.

Parser changes look good to me!

rogerburtonpatel:
Just saw this, removed old and unused versions + binary. If this version doesn't work out for some reason we can always come back to them.

PR #369, opened 2024-07-08
Add simple linker.

rogerburtonpatel:
Added the `link` function, with type ` list toplevel_entity -> list toplevel_entity -> list toplevel_entity`, and introduced it into the `TopLevel` call chain at sites like `interpreter_gen`, `model_gen_*`, etc. 

The code ended up being quite short once I reduced it. Tests and proof of linker postcondition are probably worth it. I'll work on those tomorrow. 

I made type aliases `ll_tle` for `toplevel_entity typ (block typ * list (block typ))`, and
`ll_tle_program` for `list ll_tle`. 
I think this makes the code I wrote and the code that references a variable `(prog : ll_tle_program)` cleaner. 

Lastly, the list of predefined definitions is currently empty. I'll add code to populate it (or hardcode it, depending on your preference) tomorrow.

### DIFF
--- a/src/coq/Semantics/TopLevel.v
+++ b/src/coq/Semantics/TopLevel.v
@@ -241,6 +241,66 @@ Module Type LLVMTopLevel (IS : InterpreterStack).
   
   (* TOPLEVEL Semantics  ------------------------------------------------------------------------- *)
   
+
+  (** * Linking 
+
+    We first need to link external definitions. Currently, these definitions are
+   only functions we hard-code into the environment for their usefulness-- 
+   most notably `printf`. Linking occurs at the `toplevel_definition` level. *)
+
+  Definition ll_toplevel_entity := (toplevel_entity typ (block typ * list (block typ))).
+
+  Definition ll_toplevel_entities := toplevel_entities typ 
+                                                 (block typ * list (block typ)). 
+
+  Definition PREDEFINED_FUNCTIONS : ll_toplevel_entities := []. 
+  (* NEXT: Fill with `printf`'s definition and others. *)
+
+
+  (* checks if `userdecl_n` is a name of a definition in `predefs`. *)
+  Definition userdecl_defined_in (userdecl_n : string) 
+                                 (predefs    : ll_toplevel_entities) : bool := 
+    existsb (fun predef =>
+    match predef with 
+          TLE_Definition  {| df_prototype := {| dc_name := Name predef_n |}|}
+             => userdecl_n =? predef_n
+      | _    => false
+    end) predefs. 
+
+  Definition decl_defined_in (user_tle : ll_toplevel_entity) 
+                             (predefs  : ll_toplevel_entities) : bool := 
+    match user_tle with 
+      | TLE_Declaration {| dc_name := Name n |} => userdecl_defined_in n predefs
+      | _ => false 
+    end. 
+
+  Definition defines_decl : 
+    ll_toplevel_entities -> ll_toplevel_entity -> bool := 
+    (flip decl_defined_in).
+
+  (** Importantly, the linker _removes any declaration_ from the user's
+     program that shares a name with a definition in `predefs` 
+     before joining the two programs. 
+     This is so that we can enforce our handcrafted declaration 
+     is the one referenced by their program. 
+  *)
+  Definition link  (predefs  : ll_toplevel_entities)
+                   (userprog : ll_toplevel_entities) : ll_toplevel_entities := 
+      let predefined     := (defines_decl predefs) in
+      let userprog'      := filter (negb ∘ predefined) userprog 
+        in List.app predefs userprog'. 
+
+   (* Worth a proof that linking preserves structure when there are no 
+      duplicates + upholds the linking removal postcondition?
+   
+   Lemma linking_postcondition :  
+    forall (p1 p2 : ll_toplevel_entities), 
+      not exists tl in (link p1 p2), tl' in p1 s.t. 
+      (tl is a Definition with name n /\ tl' is a Declaration with name n)
+   
+      Will write tomorrow if so.  *)
+
+
   (** * Initialization
 
     The initialization phase allocates and initializes globals, and allocates
@@ -360,7 +420,7 @@ Module Type LLVMTopLevel (IS : InterpreterStack).
      * initialize the global environment;
      * pointwise denote each function (and builtin)
      * retrieve the address of the entry point function;
-     * tie the mutually recursive know and run it starting from the
+     * tie the mutually recursive knot and run it starting from the
      * entry point
      *
      * This code should be semantically equivalent to running the following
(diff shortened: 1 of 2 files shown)


src/coq/Semantics/TopLevel.v
```diff
@@ -241,6 +241,66 @@ Module Type LLVMTopLevel (IS : InterpreterStack).
...
+
+  (** * Linking 
+
+    We first need to link external definitions. Currently, these definitions are
+   only functions we hard-code into the environment for their usefulness-- 
+   most notably `printf`. Linking occurs at the `toplevel_definition` level. *)
+
+  Definition ll_tle := (toplevel_entity typ (block typ * list (block typ))).
+
+  Definition ll_tle_program := list ll_tle.
```
Chobbes: In LLVMAst.v we have a `toplevel_entities` already:

```
Definition toplevel_entities (FnBody:Set) : Set := list (@toplevel_entity FnBody).
```

Maybe we should use that, though we would have to pass it the `(block typ * list (block typ))` type.

src/coq/Semantics/TopLevel.v
```diff
@@ -241,6 +241,66 @@ Module Type LLVMTopLevel (IS : InterpreterStack).
...
+     either side of the equation can be expanded to check for `tle` as a 
+     definition and to check for declations in `tles` by commenting in 2 lines. 
+     if we do this, we should change the contract of add_predefs to make note
+    of it. 
+  *)
+  Definition decl_defined_in (tle : ll_tle) (tles : ll_tle_program) : bool := 
+    existsb (fun tle' =>
+    match tle, tle' with 
+      | (  TLE_Declaration {| dc_name := Name n |} 
+        (* add C-O'd lines if we want to expand match *)
```
Chobbes: What does `C-O'd` mean?

src/coq/Semantics/TopLevel.v
```diff
@@ -241,6 +241,66 @@ Module Type LLVMTopLevel (IS : InterpreterStack).
...
+
+
+  (* checks if `tle` is a declaration that has a definition with the same name
+     in `tles`. 
+     either side of the equation can be expanded to check for `tle` as a 
+     definition and to check for declations in `tles` by commenting in 2 lines. 
+     if we do this, we should change the contract of add_predefs to make note
+    of it. 
+  *)
+  Definition decl_defined_in (tle : ll_tle) (tles : ll_tle_program) : bool :=
```
Chobbes: Nit-picky, but might make sense to specialize this so the `tle` parameter is just the string instead so we get less match-noise. We could even use that as a helper function to define this one.

src/coq/Semantics/TopLevel.v
```diff
@@ -241,6 +241,66 @@ Module Type LLVMTopLevel (IS : InterpreterStack).
...
+     definition and to check for declations in `tles` by commenting in 2 lines. 
+     if we do this, we should change the contract of add_predefs to make note
+    of it. 
+  *)
+  Definition decl_defined_in (tle : ll_tle) (tles : ll_tle_program) : bool := 
+    existsb (fun tle' =>
+    match tle, tle' with 
+      | (  TLE_Declaration {| dc_name := Name n |} 
+        (* add C-O'd lines if we want to expand match *)
+         (* | TLE_Definition  {| df_prototype := {| dc_name := Name n |}|} *)
```
Chobbes: Delete.

src/coq/Semantics/TopLevel.v
```diff
@@ -241,6 +241,66 @@ Module Type LLVMTopLevel (IS : InterpreterStack).
...
+        ), 
+        (  TLE_Definition  {| df_prototype := {| dc_name := Name n' |}|}
+         (* | TLE_Declaration {| dc_name := Name n' |} *)
```
Chobbes: Delete.

src/coq/Semantics/TopLevel.v
```diff
@@ -241,6 +241,66 @@ Module Type LLVMTopLevel (IS : InterpreterStack).
...
+
+  (** Importantly, the linker _removes any declaration_ from the user's
+     program that shares a name with a definition in `predefs` 
+     before joining the two programs. 
+     This is so that we can enforce our handcrafted declaration 
+     is the one referenced by their program. 
+  *)
+  Definition link  (predefs  : ll_tle_program)
+                   (userprog : ll_tle_program) : ll_tle_program := 
+      let not_predefined := (negb `o` ((flip decl_defined_in) predefs)) in
```
Chobbes: Maybe we should pull this `(negb `o` ((flip decl_defined_in) predefs))` into a separate named function for readability. Maybe a `is_defined_in` and then we can punt the `negb` into the filter.

Chobbes (changes requested):
Looks good! Left some picky comments :).

src/coq/Semantics/TopLevel.v
rogerburtonpatel: Changed, and renamed to something closer to the original. Lmk what you think. The alias is really just for type annotations below to not spill like crazy.

src/coq/Semantics/TopLevel.v
rogerburtonpatel: "Commented-Out." Deleted per below so moot now :-)

src/coq/Semantics/TopLevel.v
rogerburtonpatel: Done.

src/coq/Semantics/TopLevel.v
rogerburtonpatel: Pulled a bit out. There's a limit to what you can do since `decl_defined_in` or `is_defined_in` needs to take`predefs` as a curried argument to capture it for the call to `filter`. If we assumed that the linker was of type 
`toplevel_entities -> toplevel_entities` (type arguments excluded for brevity), and that the old first argument was _always_ the global predefined functions, then this would be an easy fix:

```
  Definition is_defined_in :  ll_tle -> bool := 
    ((flip decl_defined_in) PREDEFINED_FUNCTIONS) (* PREDEFINED_FUNCTIONS is a global variable *)
```
 That might be restrictive if we intend to build this linker out further. 

For now, pulled out what I think is the happiest medium.

src/coq/Semantics/TopLevel.v
Chobbes: Yeah, honestly I wonder if at some point we should use a proper non-empty list type instead of this tuple thing. We do have a `NonEmpty.v` in Utils already. It's not used much, though. Maybe there's an existing non-empty type in the standard library too.

rogerburtonpatel: See https://github.com/vellvm/vellvm/pull/371!

rogerburtonpatel:
Ready to merge?

PR #370, opened 2024-07-09
Ast printer vector and vaarg fix

rogerburtonpatel:
Fixed one issue with vectors and a separate one with vaargs. The vaargs example was subtle.


Chobbes (changes requested):
This looks great! Should we include the `printf-ast` file in this, though?

rogerburtonpatel:
If it's alright, I've found it helpful to keep it around for now while we decide how exactly we're going to use the definition for reference and copy-paste. Once its location is decided on, I've made a note to send it to the wind.

rogerburtonpatel:
That said, I'm very pro throwing it in `.gitignore` so others branching from `dev` won't catch a big file. I'm off my computer now and can happily do this tomorrow.

Chobbes:
> If it's alright, I've found it helpful to keep it around for now while we decide how exactly we're going to use the definition for reference and copy-paste. Once its location is decided on, I've made a note to send it to the wind.

I think it's reasonable to keep it in your directory for experiments, but it probably shouldn't be checked in to the repo. If nothing else the `print-ast` file doesn't belong in this particular PR.

Personally, I also like the idea of being able to run the vellvm interpreter with multiple `.ll` files that will be linked together, so we could have a `stdio.ll` that we parse and link in, instead of directly incorporating it into the Coq code. If we do that we'd want to check in the `.ll` file for `printf` in like a `stdlib` directory or something.

rogerburtonpatel:
Makes sense, especially agreed on wrong PR. Removed altogether.  

I think the next step for me is to build and link to `stdio.ll`. For now I think it's a good middle step to use the Coq code to get a `prinft` MWE, play with it a bit, and then build on the linker to get more real.

PR #371, opened 2024-07-10
Added nonempty list library and some proofs

rogerburtonpatel:
Per: comment abt nonempty lists in https://github.com/vellvm/vellvm/pull/369#discussion_r1671005045. 

Happy to write more proofs if they're warranted!


Chobbes (commented):
Maybe this nlist stuff should be moved out of Numeric/Coqlib eventually, and then we have NonEmpty which we should consolidate at some point.

PR #373, opened 2024-08-20
Fix the link for the ICFP'22 paper

lastland:
The link for the ICFP'22 paper no longer works. I replace it with its DOI for the PACMPL version.


Chobbes:
Good catch, thanks!

PR #377, opened 2024-11-21
Add support for opaque pointers

Ef55:
Add support for [opaque pointers](https://llvm.org/docs/OpaquePointers.html), i.e. the `ptr` type:
```
%1 = alloca i32, align 4
store i32 0, ptr %1, align 4
```

TODOs:
- [ ] Properly adapt QuickCheck code
- [x] More tests


Chobbes:
@Ef55 this looks pretty close to being ready to merge. Is there still anything left to do? It looks like there's a few cases in QC where having the opaque pointer could potentially be problematic, but at first glance how you've patched things up should work for now (because we don't generate opaque pointers).

Don't want to bump any of your other priorities, but if this is ready I think it'd be a great addition :).

Ef55:
> Is there still anything left to do?

Thanks for reminding me :). I just had forgotten that I needed to add some test files.

> It looks like there's a few cases in QC where having the opaque pointer could potentially be problematic, but at first glance how you've patched things up should work for now (because we don't generate opaque pointers).

Your analysis is right: I patched the QC files by pretending that opaque pointers don't exist. Theses files are exercised in the CI, right? That would imply that this branch passing the CI indicates that the patch is good enough (for now).

I just pushed a [test script](https://github.com/vellvm/vellvm/blob/daff0060e934d8bebf4b5f8f97da50593c6d3632/tests/opaque-ptrs/convert.sh) which can be used to update test files to use opaque pointers instead of the old pointers.
It could be used to update all tests to the new syntax, but I didn't do that. I instead used it to update some testcases in a separate directory, such that both new- and old-style pointers are tested.
I nonetheless updated all tests locally and ran vellevm on them as an experiment; everything seemed to work just fine.

@Chobbes if the QC-related files look alright to you, and with the addition of the test cases, we can merge this branch as far as I'm concerned.

Ef55:
Note: It might be useful to add notations/definitions for pointers, e.g.
```coq
(* "Typed" pointer *)
Notation tptr t := (TYPE_Pointer (Some t)).
(* Opaque pointer *)
Notation optr := (TYPE_Pointer None).
```

Chobbes:
The QC files aren't exercised by CI, but I'll double check QC before merging. We should probably add QC to CI, but... It doesn't always succeed because there are bugs in Clang / Vellvm (mostly Clang at this point, I think, because we fix the ones that we think are Vellvm's fault).

But this is great! I'll check it out and make sure it's good with QC and get it merged :). I think this feature will make it much easier to implement things like file handling (without it clang generates waaaaaay too much info about the underlying type of file pointers).

Ef55:
> I think this feature will make it much easier to implement things like file handling [..]

Great to hear! I didn't expect this change to be useful in such ways.

I unfortunately noticed that my script is not complete enough (doesn't handle pointed arrays, which can't be properly parsed using regexes); I'll leave the script as it is (I don't want to use perl regexes to solve that issue), but manually fix the test cases; I'll push as soon as the CI is done.

PR #384, opened 2025-06-26
Parse landingpad

Zdancewic:
Added parsing support for `invoke` instructions.  This was a bit more annoying than one might think because `invoke` is now the only "terminator" that can define a local ID.

Also added some test cases for parsing.  These compile and run with clang, but they don't throw/catch exceptions, so they currently only execute the "normal" control flow path.

### DIFF
--- a/src/dune-project
+++ b/src/dune-project
@@ -1,2 +1,2 @@
-(lang dune 3.0)
-(using menhir 2.0)
+(lang dune 3.14)
+(using menhir 3.0)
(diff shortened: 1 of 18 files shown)


Chobbes:
This looks good to me, I suppose the tests that are failing are failing because invoke isn't currently handled?

Zdancewic:
Yes, that's right -- I added two executable test cases that use `invoke` and `landingpad` -- they compile and run with clang, but not (yet) in our semantics.  They don't actually test throwing/catching exceptions yet, though.

Chobbes:
Okay. I think it's fine to merge if that's convenient. We could also wait until we can pass the tests, or maybe mark the tests as expected to fail for now. Not a huge deal either way.

PR #388, opened 2025-07-18
parse operand bundles

Zdancewic:
Only lightly tested, but we now pass the old failing test case that failed to to operand bundles.


Chobbes:
Awesome! Assuming CI passes I'll get this merged in.

Chobbes:
I think opam did a renaming of things for Rocq that maybe broke some dependencies on opam. Maybe it's a matter of changing some coq-* stuff to rocq-* stuff in the opam files.

https://github.com/math-comp/hierarchy-builder/tree/master

This seems to be what the build breaks on. I'll see if updating the dependencies helps.

Zdancewic:
Hmmm... What is the status, then?  It looks like the problem is the renaming of `coq-hierarchy-builder` to `rocq-hierarchy-builder`.  Maybe triggered by the change from `coq-elpi.2.5.0` to `rocq-elpi.3.0`?

PR #409, opened 2025-12-23
add support for the disjoint keyword for or

Zdancewic:
Still missing some of the proofs (grep for `admit`).


Zdancewic:
We'll also have to regenerate the definition of `printf` for this one.

PR #414, opened 2026-01-30
metadata support on atomic... instructions

Ptival:
e.g.
```
%0 = atomicrmw sub ptr %_6.i.i.i, i64 1 release, align 8, !noalias !1903
```

Not sure whether you have a test suite or a set of regression tests you'd want to change, let me know.

This simple fix seems to work enough for me. I run into another error next, but I think it's unrelated:
```
Fatal error: exception ParseUtil.InvalidAnonymousId("Unexpected sequential id: expected >= 3 but found 2")
```

### DIFF
--- a/src/ml/libvellvm/llvm_parser.mly
+++ b/src/ml/libvellvm/llvm_parser.mly
@@ -1665,6 +1665,7 @@ instr:
   | KW_ATOMICCMPXCHG c_weak=KW_WEAK? c_volatile=KW_VOLATILE?
     c_ptr=texp COMMA c_cmp=texp COMMA c_new=texp ss=syncscope?
     c_success_ordering=ordering c_failure_ordering=ordering c_align=comma_align?
+    md=instr_metadata
     { let f_info = metadata_file_info $startpos $endpos in     
       let c_syncscope =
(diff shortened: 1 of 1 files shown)


Chobbes:
@Ptival there is a test suite in `tests/` the README has some instructions for how to add test cases (not 100% sure how up to date it is). I think there's some subtlety with how tests run. I think if you add a file and add asserts it will run automatically, otherwise it has to be manually added to an older bespoke test harness in the ocaml code. Maybe @Zdancewic can confirm.

Looks like Steve just barely beat you to the punch on this PR, which is a little unfortunate (but it's always good to have more people able to modify the parser). Another plus is it looks like Steve added a test case, so there's an example of adding a test case for this kind of change:

https://github.com/vellvm/vellvm/commit/fbdf2e15777ddca0d52f10dbf7bfc94e39e24415

I guess ideally we should probably have one for atomiccmpxchg too.

I'm going to close this PR for now, because it looks like equivalent changes have already been incorporated, but please correct me if I'm mistaken, and thank you for your effort!

Ptival:
All good, thanks for the info!

PR #422, opened 2026-03-26
Chop up LangRefine into a bunch of different files.

Chobbes:
This is a large refactor of LangRefine, splitting everything into many files. This should build, but the explosion of module types, additional parallelism, and possibly the reordering of hints in the ORUTT database have caused the time and memory necessary to compile vellvm to skyrocket (to the point where CI seems to run out of memory).


Zdancewic:
This pull request is obsolete -- breaking up the development in this way increased compilation time.

PR #429, opened 2026-05-28
Major refactoring laying ground to a port to ctrees

YaZko:
** DISCLAIMER : This is not meant to be merged in `dev` yet**

This PR performs a set of changes aiming at simplifying and minimizing the repo before porting the project to ctrees.

The core changes go as follows:

## Removal of undef

The management of symbolic values (`uvalues`) to perform concretization by need is a humongous source of complication, for a feature that is being actively removed from llvm IR. We therefore chose to be optimistic and look forward.

Consequences:
- No more `uvalues`, we pass `dvalues` directly around everywhere
- A lot of semantics previously inlined in `concretization` is now leaving in their own files, gathered in `Operations.v`, and called in `Denotation.v`.
- The `pick` events are for now replaced by a simpler `draw` event drawing a value at a dynamic types, and exclusively used for implementing `freeze`
- Bytes are initialized to `poison` instead of `undef`. We will want to introduce poisoned bits following ongoing changes on the llvm side

## Minimizing the number of monads involved

To simplify things, we use the following monads:
- EOB (that needs to be renamed) in `Utils/EOB.v` is a trusty simple monad for error, oom, and ub. It is used for every computation that is known to be finite outside of the memory model: we don't try anymore to nitpick if it could just live in the error monad, or an error or ub one, and so on and so forth...
- The memory model lives in a monad `memM` defined in `Semantics/Params/Memory.v`, a little free monad supporting the same three failures as `EOB`, as well as stateful `get/put`, and two operations to get a fresh provenance and a new raw address. We have a simple embedding of `EOB` into `memM`.

## Complete refactoring 

Removed all usage of modules, and replaced them by type classes.
- In `Semantics/Params` are a set of files containing classes, usually split in two between operations and theory. These attempt to minimally mirror the previous Module Type, although the matching is tricky to do since these interface were polluted by inlined implementation for sharing.
- Note: a swap was performed so that the `Addr` class is aware of the Provenance one rather than the other way around
- Implementations are provided in `Semantics/Params/VellvmImplem`. The elementary parameters (i.e., all but the memory model) are implemented parametrically in an instance of `IPtr`. This specific class is provided two instances, a finite one and an infinite one.
- All these elementary parameters are packaged into a `Params` class, of which an instance is therefore exposed parametrically in an instance of `IPtr`, so that other parts of the development can simply assume locally an instance of `Params`.

### The case of the memory model

- In `Semantics/Params/Memory.v`, we expose the interface for the state and its primitive operations. This interface is extracted from the previous Module Type, but should be reconsidered: the primitives are very sensible, but the set of pieces of the state that are exposed seem completely random at the moment.
- In `Semantics/Params/VellvmImplem/Memory.v`, we define two instances. First an implementation of the high level operations of the memory model parametrically in an instance of the primitives. This culminate into two handlers for `MemE` and `IntrinsicE` *into memM*.
- Second a concrete instance of the primitives, implementing the model.
- We now can turn our attention to `Handlers/Memory.v`. We need to get now a handler from `MemE + IntrinsicE` *into stateT (itree _)*. We do so by composing the `memM` handlers with an interpretation of `memM` into such stateful itrees.
- *This is the place where we decide if we work on the model or the executable memory model*. One interpretation gives deterministic implementations to `fresh_provenance` and to `next_key` while the other keeps all options possible. The intent is to implement the latter as a ctree choice node: it is therefore not implemented yet.
- Note: the same dual implementation happens for handling `Draw`

### Top level and summing up

For now we only rebuild and extract the Infinite interpreter. The four options are however meant to be built by picking either the finite or infinite instance of `IPtr`, and feeding either the deterministic or non-deterministic implementation of `memM` into `stateful ctrees`.

Disclaimer: a lot of things have been cut off for the development, and are intended to be rebrought in on demand.


YaZko:
This PR was closed as completed by reaching step 1: a first reasonably complete port of the interpreter has been reached. We have upgraded it to the `main` branch, and consider it as the new basis of the modern infrastructure for the project.

From there, ongoing perspectives include:

1. in the absence of undef bytes, uninitialized memory is made of poison bytes. Reading a sequence of bytes including one such lead to a poison value. However, we now implement `load` such that it freezes the read value, which seems to be enough to pass our internal tests. We now want to consider the implementation of byte types and poison bits in the style of the PLDI'26 paper.
2. although the internal tests are passing, we need to take stocks on the Rust-compiled ongoing investigations and fix the interpreter accordingly.
3. with the interpreter getting mature, we now want to replug the model as well. We intent to do so using ctrees, with a monad morphism from `memM` to stateful ctrees.
4. we progressively rebuild the infrastructure for proving a refinement from the infinite to the finite memory model---see the `I2F` branch.

Issue #428, opened 2026-05-27
Potential bug around floating point computations


Ptival:
I think I'm seeing a bug in how Vellvm handles some floating point multiplication. It's hard to be certain, so I'll give all the information I gathered.

Related to #427 , I'm using 32-bit hex syntax in the assertions, which seems like what Vellvm handles least confusingly. Also using non-hex constants to show that the problem seems to be in the computation itself:
```llvm
; ASSERT EQ: float 0x3a800000 = call float @frac()
; ASSERT EQ: float 0.0009765625 = call float @frac()
define float @frac() {
  %frac = fdiv float 1.0, 1024.0
  ret float %frac
}

; ASSERT EQ: float 0x39638e39 = call float @main()
; ASSERT EQ: float 0.00021701389050576836 = call float @main()
define float @main() {
  %frac = call float @frac()
  %ret = fmul float %frac, 0x3FCC71C720000000 ; = 0x3e638e39
  ret float %ret
}
```

The assertions for `@frac` succeed, indicating the `fdiv` computation happens as expected.

But for the `fmul` in `@main`, the result seems completely incorrect.
The correct result should be `0x39638e39` (or `0x3f2c71c720000000`, or `0.00021701389050576836`).

The actual result computed by Vellvm seems to be `0x9b000000`/`0xbb60000000000000`/`-1.0587911840678754e-22`.

Compiled to assembly and checked via gdb, the result is `0x39638e39`.

Chobbes:
The issue almost certainly has to do with how we handle floating point literals (particularly doubles as floats?) My best guess is that the issue occurs here in `VellvmFloats.v`, but I'm not as familiar with the code. It seems like this code is expecting the hex to yield a 32-bit integer, but since that's not the case it's probably passing `Bits.b32_of_bits` a massive integer and it's spitting out garbage.

I'm sure it's in LangRef, but I'm unsure how this case should be handled. Maybe it gets converted from a double with a specific rounding mode?

@Zdancewic probably has a better idea than me. My assessment of this might be very wrong---I've always been a little confused about what's going on with the floating point literals.

``` rocq
(* Converting hexadecimal to float is much easier. *)
Definition hexadecimal_uint_to_float32 (h:Hexadecimal.uint) : float32 :=
  Bits.b32_of_bits (BinInt.Z.of_hex_uint h).


Definition float32_of_float_syntax (fs:float_syntax) : option float32 :=
  match fs with
[...]

Issue #427, opened 2026-05-21
Potential bug around floating point constants


Ptival:
While debugging some floating point problems, I was able to produce this C program:

```c
int main() {
    static const float floats[2] = { 813694976, 2147483648 };
    return (floats[0] == floats[1]);
}
```

producing this LLVM IR:

```llvm
target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-i128:128-f80:128-n8:16:32:64-S128"
target triple = "x86_64-unknown-linux-gnu"

; ASSERT EQ: i32 0 = call i32 @main()

@main.floats = internal constant [2 x float] [float 0x41C8400000000000, float 0x41E0000000000000], align 4

define dso_local i32 @main() {
  %1 = alloca i32, align 4
  store i32 0, ptr %1, align 4
  %2 = load float, ptr @main.floats, align 4
  %3 = load float, ptr getelementptr inbounds ([2 x float], ptr @main.floats, i64 0, i64 1), align 4
  %4 = fcmp oeq float %2, %3
  %5 = zext i1 %4 to i32
  ret i32 %5
}
```

The program returns 0, as the equality test fails. However, Vellvm says:
```
FAILED - i32 0 = @main()
           ERROR: dvalue comparison for equality failed:
got:
        i32 1
asserted:
        i32 0
```

So I feel like there is possibly something funky going on in the handling of floats.

YaZko:
Tested in main: vellvm appears to comply properly now. Added the program to the test-suite.

PR #430, opened 2026-05-29
fix the parsing of 64-bit float hex constants as 32-bit float

Ptival:
This is a proposed fix for #428 and likely also #427.

Now, the criterion for representability is fairly involved, so an alternative if you don't want to bring in this complexity would be to do some roundtrip conversion.

If that would be the preferred solution, I can make a PR with the roundtrip version instead.


Zdancewic:
The code from this pull request is now reflected on the new `main` branch.

PR #447, opened 2026-07-13
Memoization of wordsize

YaZko:
A little silly, but memoizing the wordsizes turn out to give a good 20% speed up on most of the performance test cases from #446 . Kinda ugly but probably worth it.


Zdancewic (approved):
This looks great to me!

PR #450, opened 2026-07-15
Avl locals globals

YaZko:
This contains two performance patches:
- First I simply swapped the association map used for locals and globals in favour of AVL, as already used in the memory.
- But then it did not reduce as much the benchmark stressing locals, so I asked Claude its opinion, and it found something I did not see at all: `map_monad_` was inefficient because it collected the result before discarding it. Swapping it for an implementation that discards results on the go makes a much more drastic change. 

Namely, with of course a very non-scientific measuring on my own machine and no repetition, we have for the locals-chain stress test:
- base: 9.05 s
- avl maps for locals: 6.53 s
- avl maps for locals and optimized map_monad_: 0.24 s


YaZko:
Note: I did let Claude write the commit to separate both features cleanly. This implied that it put itself as collaborator. We may want to remove this, I am not sure.

Zdancewic (approved):
Looks fine to me!

PR #451, opened 2026-07-16
Cfg avl

YaZko:
We were still representing cfgs as lists of blocks, such that lookups during jumps were linear.

Rather than changing cfgs themselves, we keep them as lists on the syntax side, but introduce an AVL in the denotation/interpreter. A couple of performance tests illustrate the impact.

Contrary to what I remembered we already had AVLs for mcfgs so nothing to be done there, but added a perf test for it as well.


YaZko:
More broadly, I used Claude to identify clear scaling issues in the interpreter. This new set of performance tests track them.

Zdancewic:
I think that this looks great.  I think that is also addresses most of the relevant outstanding items in pull request #389 ?

YaZko:
> I think that this looks great. I think that is also addresses most of the relevant outstanding items in pull request #389 ?

Arguably #389 also was hoping to benefit from stdpp's proof automation, but it solves the efficiency concerns from this PR definitely! I had not thought of it but I agree I wouldn't mind closing #389 for now.

PR #454, opened 2026-07-16
Conversion refactor

Zdancewic:
This pull request factors the conversion operations into:

- "pure" conversions, which don't interact with the memory model
- "impure" conversions, which might interact with the memory model
- `bitcast`, which is special because it can't be vectorized and can convert vector to non-vector types

It also adds syntactic support for the new `ptrtoaddr` and `addrspacecast` instructions, which are included in the "impure" conversions.

For this branch, `ptrtoint` and `ptrtoaddr` behave identically (and are actually pure).

It wasn't too bad to port the `I2F` proofs to this way of doing the conversions.  One proof (`I2F_convert_pure_base`) is a bit slow--I think it always was--but over all it seems pretty good.


YaZko:
Yes I confirm this proof was already the slowest, it generates 100 of cases. I haven't tried too hard to speed it up though, we could take a look.

Zdancewic:
The whole development still builds in under a minute on my machine, so that proof isn't too bad.

Issue #434, opened 2026-06-24
Suggestion: local tracing auto-stepping debugger mode


Ptival:
When debugging issues, I'm frequently manually stepping the debugger, then printing locals, then stepping some more and printing locals again. But the locals context is monotonically increasing, and I'm only interested in the new locals.

So I'm thinking it'd be nice to have something that is a bit like the debug mode, where it would automatically step through each instruction, indicating what instructions it's at, and print out locals as they get assigned.

Essentially, something a bit like what `-debug` with `-test-file` is printing, but with locals interleaved as they become assigned, e.g.:
```llvm
[../Tractor-Clear/docker/tractor-test-corpus/crisp-linked/pow43-linked.ll:52.6-52.34]
[../Tractor-Clear/docker/tractor-test-corpus/crisp-linked/pow43-linked.ll:216.17-216.36]
x -> i32 1025
[../Tractor-Clear/docker/tractor-test-corpus/crisp-linked/pow43-linked.ll:217.2-217.41]
x.dbg.spill -> ptr (1727, Some [109])
[../Tractor-Clear/docker/tractor-test-corpus/crisp-linked/pow43-linked.ll:219.6-219.84]
[../Tractor-Clear/docker/tractor-test-corpus/crisp-linked/pow43-linked.ll:226.20-226.41]
0 -> i32 1025
[../Tractor-Clear/docker/tractor-test-corpus/crisp-linked/pow43-linked.ll:227.20-227.39]
frac.dbg.spill -> ptr (1731, Some [111])
[../Tractor-Clear/docker/tractor-test-corpus/crisp-linked/pow43-linked.ll:228.10-228.29]
sign.dbg.spill -> ptr (1735, Some [112])
...
```

PR #455, opened 2026-07-16
tentative tracing version of step

Ptival:
This was my attempt at https://github.com/vellvm/vellvm/issues/434

I needed something like this for a program that I needed to step thousands of times before reaching the problem.

I think it could be improved by people more familiar with the stack frames, and more familiar with OCaml, but at least I'm now using this on my side.

### DIFF
--- a/src/ml/debugger.ml
+++ b/src/ml/debugger.ml
@@ -54,13 +54,58 @@ let debug_raw_step m =
      Either.Left x
   | res -> res)
 
+(**
+Takes in two lists, and returns a triple:
+- pairs of elements with matching indices from l1 and l2
+- either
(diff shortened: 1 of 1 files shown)


Ptival:
It's not printing as much as I need, I will try to improve it.

EDIT: Oh I now see that the RawIdMaps are AVL trees, so I need to actually perform a set difference between the sets of keys to find the new keys. Will update.

Ptival:
I have fixed the AVL problem.

My next problem is that sometimes a local gets overwritten in a loop.
So I might implement detection that the value of a local was updated, if it's not too painful.

EDIT: It was easy to implement, and it's working well for me! PR updated.

PR #459, opened 2026-07-20
Fuse interps

YaZko:
I've been meaning to test this for a long time, and it's easy to do now we have some little profiling target: does this layered interpreter approach costs much in run time? This branch fuses the three stateful passes (and the `draw` one) together into one single pass and compare the interpreters.

## Design choice

Slight design choice: here I keep the handlers identical, in their "small" monad, and add a widget when combining them to embed them into the larger stateful monad. That seems reasonable, this way they don't have to be aware of the whole state, and shouldn't be much overhead to reason about.

Overall I actually think it would also be easier to reason about: we lose the mythical "ability to prove a transformation pass that would depend on different registers but would still have the same memory trace without introducing the memory" that we never exercised, but we would also not have to painful go through layers of interpreters, painfully commuting with each of them when building the theory.

## Results

On this `fuse-interps` branch:
```
test                      seconds  assertions
alloca-churn.ll              1.22  Passed: 1/1
alloca-large.ll              2.39  Passed: 1/1
block-jumps.ll               3.19  Passed: 1/1
calls-fib.ll                 3.46  Passed: 1/1
calls-large-fn.ll            1.55  Passed: 1/1
calls-many-fns.ll            2.39  Passed: 1/1
global-init.ll               1.02  Passed: 1/1
iptr-roundtrip.ll            1.86  Passed: 1/1
locals-chain.ll              0.19  Passed: 1/1
loop-phi-arith.ll            2.98  Passed: 1/1
mem-aggregate.ll             1.54  Passed: 1/1
mem-scan.ll                  2.73  Passed: 1/1
memcpy-chunk.ll              3.60  Passed: 1/1
memset-chunk.ll              2.15  Passed: 1/1
switch-cases.ll              3.19  Passed: 1/1
undef-pick.ll                1.59  Passed: 1/1
vector-ops.ll                1.27  Passed: 1/1
wide-arith.ll                2.51  Passed: 2/2
```

On the `switch-opt` branch:
```
test                      seconds  assertions
alloca-churn.ll              1.63  Passed: 1/1
alloca-large.ll              2.39  Passed: 1/1
block-jumps.ll               5.21  Passed: 1/1
calls-fib.ll                 5.39  Passed: 1/1
calls-large-fn.ll            2.51  Passed: 1/1
calls-many-fns.ll            3.39  Passed: 1/1
global-init.ll               1.01  Passed: 1/1
iptr-roundtrip.ll            2.75  Passed: 1/1
locals-chain.ll              0.21  Passed: 1/1
loop-phi-arith.ll            4.83  Passed: 1/1
mem-aggregate.ll             1.91  Passed: 1/1
mem-scan.ll                  3.62  Passed: 1/1
memcpy-chunk.ll              3.85  Passed: 1/1
memset-chunk.ll              2.20  Passed: 1/1
switch-cases.ll              3.39  Passed: 1/1
undef-pick.ll                2.32  Passed: 1/1
vector-ops.ll                1.92  Passed: 1/1
wide-arith.ll                3.75  Passed: 2/2
```

So we seem to gain a good 30% performance (eye-balling rapidly) on lengthy tests!


YaZko:
Thinking a bit more, we need to think slightly forwards: right now, intrinsics are not fused with everything else because they emit (memory) events. This suggests two things:

- If we care a lot about performances, we could revert the change where I pushed some of the intrinsics to the first handler by allowing it to emit events, and instead re-host them inside of the memory model
- Conversely we need to be wary that if the model were to emit events that must be caught lower (I can't think of any other than `draw`), then it should be reflected on the interpreter side as well to not break the structure of the I2F proof (although the part relating interpreters is not written yet).

PR #462, opened 2026-07-25
Complete Infinite to Finite proof of refinement for the interpreter

YaZko:
This completes the proof of refinement that was previously done up to the denotation.

Main new ingredients are:
- A refinement at the level of `memM`, so that most of the proof for the memory model can be shared between interpreter and future model
- Refinement for every handler
- Lifting of refinements through interpreters
- Refinement for the top level, and putchar/puts in particular

The new main theorem is `I2F_interpreter_param`.

Claude has been used in several places to develop this proof. It's been good at generalizing the infrastructure and filling up the repetitive work (such as covering every intrinsics for instance). It's always a delicate tool to use, it both generalizes invariants to various constructions very efficiently and rapidly, and gets tripped on an implicit argument or a syntactically incorrect invokation of a tactic for ungodly amount of time.

Note: printing dependencies for `I2F_interpreter_param` shows off the use of `ClassicalDedekindReals.sig_not_dec` and `ClassicalDedekindReals.sig_forall_dec`. I don't particularly mind but I don't know where this dependency comes from, I need to look into it.


YaZko:
Ah the axioms come from Flocq, no problem.

PR #472, opened 2026-07-31
Fix to float_to_float32 against NaN

YaZko:
LangRef specifies for bitcast:

> The ‘bitcast’ instruction converts value to type ty2. It is always a no-op cast because no bits change with this conversion. The conversion is done as if the value had been stored to memory and read back as type ty2.

This was not respected for NaNs: CompCert's inherited `float_to_float32` would both accept too many representable values, and systematically convert any valid NaN to the same one.

I am not sure if/when we would actually observe this subtle difference, but it complies with clang (better at least) this way, added a bunch of tests by Claude with it.

### DIFF
--- a/src/rocq/Numeric/Floats.v
+++ b/src/rocq/Numeric/Floats.v
@@ -38,8 +38,13 @@ Definition is_float_representable_as_float32 (f : binary_float 53 1024) : bool :
   match f with
   | B754_zero _ => true
   | B754_infinity _ => true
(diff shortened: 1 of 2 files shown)


YaZko:
As a note: I initially pushed the new logic to VellvmFloat to avoid editing the initially inherited from CompCert float file, but the previous patch to the float semantics already patched float directly.

It's probably fine, but otherwise we'll have to do some surgery in the future if we want to explicitly separate the inherited code from the Vellvm specific one.

Issue #473, opened 2026-07-31
On valid floating point literals


YaZko:
There seems to be a decrepancy between clang and langref as far as I understand.

Namely, langref suggests that a literal should only be accepted if it can be parsed lossless:

>Value of the floating-point number if bitcast to an integer. The number must have exactly as many hexadecimal digits as is necessary for the size of the floating-point number.

However clang seems to rather take `double` as its standard of precision, allowing to be lossful at it, but then requiring bitcasts to the type of interest to be lossless _from there_. I.e., although 1.3 can never be represented, it accepts:

```
define i64 @inexact_simple_double() {
  %a = bitcast double 1.3 to i64
  ret i64 %a
}
```

But rejects:

```
define i32 @inexact_simple() {
  %a = bitcast float 1.3 to i32
  ret i32 %a
}
```

PR #474, opened 2026-07-31
Possible patch for parsing of floats, debatable

YaZko:
This patch would attempt to comply with clang w.r.t. #473 by removing @Zdancewic 's function to parse at float32, and replace it by parsing at double and converting to i32.

However since it's unclear to me that it's the behavior specified in langref, we might want to probe llvm people before merging.

### DIFF
--- a/src/rocq/Semantics/VellvmFloats.v
+++ b/src/rocq/Semantics/VellvmFloats.v
@@ -52,10 +52,7 @@ Class VFloat FP : Type :=
    Bparse 24 128 __ __ base intPart expPart.
  *) 
 
-(* Converts a syntactic floating point value to a float32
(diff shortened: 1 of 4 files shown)


YaZko:
Asked the question here: https://discourse.llvm.org/t/valid-floating-point-literals-and-langref/91439

Makes me wonder how it works for f128 too actually...

YaZko:
So it appears things precisely changed very recently, landing in the incoming next release: https://github.com/llvm/llvm-project/pull/190641

Which implements this RFC: https://discourse.llvm.org/t/rfc-floating-point-literals-in-llvm-ir/82974

But I am still confused about precision: the RFC says it intends to enforce the exact precision rule, but the answer in my thread says that they know accept any precision of literals? Not sure for now will need to dig more precisely.

PR #475, opened 2026-08-25
fix buggy preferred alignment comparison

Ptival:
I ran into a couple of alignment computation problems, the first of which is this set of comparisons that seem to accidentally use `dt1` twice instead of `dt2`.


Ptival:
Here is a test file that fails before this fix and succeeds with it:

```ll
; {i8, i32, i8}: 1 + 3 pad + 4 + 1 + 3 tail pad = 12
; ASSERT EQ: i64 12 = call i64 @size_i8_i32_i8()
define i64 @size_i8_i32_i8() {
  %p = alloca {i8, i32, i8}
  %f = getelementptr {i8, i32, i8}, ptr %p, i32 1
  %a = ptrtoint ptr %p to i64
  %b = ptrtoint ptr %f to i64
  %d = sub i64 %b, %a
  ret i64 %d
}

@fmt = private unnamed_addr constant [25 x i8] c"size_i8_i32_i8() = %lld\0A\00"

declare i32 @printf(ptr, ...)

define i32 @main() {
  %v = call i64 @size_i8_i32_i8()
  ret i32 0
}
```

Running this should print:
```
size_i8_i32_i8() = 12
```

Zdancewic (commented):
This looks great!  Thanks.

PR #477, opened 2026-08-26
fix alignment computations for aggregates

Ptival:
This fixes multiple problems with the pre-existing alignment computations:
- packedness was previously ignored,
- struct alignment was hardcoded to 8,
- same for array alignment.

I tried to save a quote from LLVM LangRef or the System V ABI next to each fix to explain why the computation should be what it is in the fix.

Specifically, for array alignment, the ABI adds an exception that I didn't bother encoding.

Here is one test for the non-packed struct case that used to fail and now succeeds:
```llvm
; ASSERT EQ: i64 3 = call i64 @size_nested_i8()
define i64 @size_nested_i8() {
  %p = alloca {i8, {i8, i8}}
  %f = getelementptr {i8, {i8, i8}}, ptr %p, i32 1
  %a = ptrtoint ptr %p to i64
  %b = ptrtoint ptr %f to i64
  %d = sub i64 %b, %a
  ret i64 %d
}

@fmt = private unnamed_addr constant [25 x i8] c"size_nested_i8() = %lld\0A\00"

declare i32 @printf(ptr, ...)

define i32 @main() {
  %v = call i64 @size_nested_i8()
  ret i32 0
}
```


Zdancewic (approved):
This also looks great!  Thanks.

